PR #7, opened 2014-12-08
Add CMake build files for Legion

chuckatkins:
Still needs some work to get the CUDA support building but works for shared and GASNet.


chuckatkins:
Will refactor for new code

PR #8, opened 2015-02-03
Fixed typo in runtime

willcrichton:
This typo prevented the runtime from building (and subsequently the entire application), so I think it's a good fix!


lightsighter:
Thanks.

PR #12, opened 2015-06-19
Eliminate unused variable warnings when compiling with -DNDEBUG

gsjaardema:


streichler:
Greg,
  Out of curiosity, how much performance improvement are you seeing with -DNDEBUG?
(I assume that's why you're doing it?)

PR #30, opened 2015-08-07
Fix DMA/master merge, change priorityxferdesqueue comparitor 

gshipman:
1) Fix merge issue where HDFMemory locks got left out
2) Change hdf_metadata to hdf_metadata_vec for clarity
3) Implement fix for PriorityXferDesQueue, reverting back to std::set and using a comparitor that provides unique keys and total ordering

### DIFF
--- a/runtime/channel.cc
+++ b/runtime/channel.cc
@@ -880,7 +880,6 @@ namespace LegionRuntime {
           switch (kind) {
             case XferDes::XFER_HDF_READ:
             {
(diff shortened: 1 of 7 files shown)


streichler:
Galen/Zhihao,
  Can we delete the dma-set-comparitor branch, or does it still serve some purpose?

gshipman:
Yes, this can be deleted.
Will also push the relevant bits from dma-merge to dma and then we can drop that one as well.
- Galen

PR #54, opened 2015-09-08
Enable building with USE_HDF and not USE_DISK on Darwin 

gshipman:
Signed-off-by: Galen Shipman gshipman@lanl.gov

### DIFF
--- a/runtime/channel.cc
+++ b/runtime/channel.cc
@@ -40,13 +40,17 @@ namespace LegionRuntime {
         return syscall(__NR_io_getevents, ctx, min_nr, max_nr, events, timeout);
       }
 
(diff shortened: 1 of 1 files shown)


streichler:
Assigned to Zhihao.  The build failures exist in the main dma branch too - the Realm tests
use a slightly different makefile and something isn't set right.

streichler:
Assigned back to Galen.  I fixed the Realm test-related build break, which caused a merge conflict
with this change (or maybe even just fixed it a different way)?

gshipman:
Sean, your change works as well, leaving XferDesQueue and ChannelManager enables building with USE_HDF without USE_DISK.

PR #82, opened 2015-10-22
Add hwloc hardware detection to realm/threads.cc.

zard49:
to use set the following environment variables:
  HWLOC     top level install location of hwloc (e.g. /usr/local)
  USE_HWLOC set to 1 to use hwloc rather that linux specific hardware detection


streichler:
Thanks again for this Dean.  As soon as Travis gives us the green light, I'll merge it in.

PR #97, opened 2015-11-10
Cosmetics

gentryx:
Hey there, nothing special here, just cleaning up a bit.


streichler:
These look fine, but they'll need to be applied to the master branch instead of stable.
I'll do that in a bit.

(We never actually delete the ReductionOpUntyped's, but if we did, we'd want this fix.)

gentryx:
Check. I'll refile against master. Sorry about that.

PR #109, opened 2015-12-02
Change clean rule to have double ::

gsjaardema:
If the clean rule is changed from clean: to clean::, then users can add a clean:: rule to their projects to clean up additional files that are not cleaned by the default legion clean rule.  

An example would be cleaning up test result output.


streichler:
"Double-colon rules are somewhat obscure and not often very useful", except in this case.  :)

PR #111, opened 2015-12-03
Permit application makefile to override USE_LIBDL

gsjaardema:
The USE_LIBDL option was not using the conditional variable assignment (?=) which most/all other options in the runtime.mk are using.

Also fix spelling error in comment


lightsighter:
Can you resubmit this pull request on the master branch?  We generally don't commit to stable branch and instead push changes to the master branch and use the stable branch to capture snapshots.

gsjaardema:
Sorry, forgot to change branch to master before creating pull request.  Resubmitted on master branch.

PR #115, opened 2015-12-13
November/December update of stable branch

streichler:
Just one update for these two months.  Next update planned for early January.


elliottslaughter:
Can we merge stable into master too? It would be nice if stable were an ancestor of master.

PR #116, opened 2015-12-14
Serdez

jiazhihao:
Pull request for merging serdez branch back to master.


streichler:
Looks like there's a bit of weirdness in the merge:
https://github.com/StanfordLegion/legion/blob/serdez/runtime/legion/runtime.h#L2156-L2157
In general, seeing "serdez" and "op" in the same line of code looks like an entanglement of the serialization stuff and the reduction op stuff from which it was copied.

lightsighter:
How did we end up with two of the same thing?

lightsighter:
I remembered the reason for the additional table so we are all good.  Merging this pull request.

PR #134, opened 2016-03-03
Added missing sys/types.h header to fix compile error on OSX

sdalton1:
Added types.h to fix error

```shell
activemsg.h:843:9 error: 'off_t' has not been declared
```

### DIFF
--- a/runtime/activemsg.h
+++ b/runtime/activemsg.h
@@ -24,6 +24,8 @@
 #include <stdlib.h>
 #include <vector>
 
+#include <sys/types.h>
+
     enum ActiveMessageIDs {
       FIRST_AVAILABLE = 140,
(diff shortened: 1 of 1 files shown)


streichler:
This should be fine, but out of curiosity, what does your compile command line look like?  We've got lots of folks building on OSX without hitting this.

sdalton1:
Hey Sean,
Here are my specs
10.10.5 (Yosemite)
Target: x86_64-apple-darwin11.4.2
gcc version 4.8.5 (GCC)

I'm using the standard Makefiles in the examples directory to build.

streichler:
The Travis errors appear to be unrelated (and after where this would have broken any of the various compilation configurations), so let's merge it in and see what happens... :)

PR #135, opened 2016-03-14
dma: informative error message for remote serdez

apoms:
Adds several error messages indicating why the remote serdez operation failed. This came up because the objects I was serializing were frequently larger than the default gasnet LMB size.

The cases added are based on the reasons why the  `assert(cur_size > 0)` would trigger which are:

1) The do_remote_serdez operation was called with a zero count, meaning no objects would be serialized and thus the size of the buffer would be zero.

2) The custom serdez function which was invoked returned a size greater than the max buffer size (the gasnet LMB) and thus could not be serialized.

3) All of the fields returned a size of zero when serializing and thus the output buffer was empty.

### DIFF
--- a/runtime/realm/mem_impl.cc
+++ b/runtime/realm/mem_impl.cc
@@ -1737,25 +1737,51 @@ namespace Realm {
       char* buffer_start = (char*) malloc(max_xfer_size);
       const char *pos = (const char *)data;
       unsigned xfers = 0;
+
+      size_t element_size = 0;
       while (count > 0) {
         size_t cur_size = 0;
         size_t cur_count = 0;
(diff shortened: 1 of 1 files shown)


streichler:
Thanks - I'll take a look.  My initial thinking is that cases 1 and 3 are ones that the dma system should handle correctly, so require changes elsewhere.  Case 2 is hard to fix right now, but perhaps
easier with Zhihao's big dma change.  In the cases where you were exceeding the LMB size, could
you have placed an upper bound on the serialized size of an element, or can they be arbitrarily
large?

streichler:
This has (finally) been rebased into master as commit 5269c67.

PR #156, opened 2016-08-04
Processor sets

zard49:
Support for launching Kokkos or OpenMP tasks from inside of legion. There were 2 main changes:

1) Add a new processor type to realm to handle "sets" of processors.

2) Update the default mapper to use these ProcessorSets.


lightsighter:
I'm ok with merging this if @streichler is.  We should probably merge into the master branch though.

streichler:
@zard49: yes, can you switch this pull request to the master branch?  (I think github just added that functionality.)

zard49:
It looks like it is already setup for master, what am I missing?

streichler:
I guess it is.  I was just looking at mike's comment above mine.  :)
That said, can you close this and open a new pull request so that Travis gets another try to test it?
(The original pull request came along when we had brokenness in the tree, and Travis isn't smart enough
to re-try.)

PR #169, opened 2016-09-04
Wrapper Mapper

rajshah11:
[Wrapper Mapper](https://github.com/rajshah11/legion/blob/wrappermappernew/runtime/mappers/wrapper_mapper.cc) which wraps a Legion mapper, so that the user can print information or stop program execution when important properties of a task are set (properties set by select_task_options()).
I have added a brief explanation on how to use it at the start of the [header](https://github.com/rajshah11/legion/blob/wrappermappernew/runtime/mappers/wrapper_mapper.h) file. Also, added a runtime parameter -hl:wrapper to wrap all mappers by default.


lightsighter:
Looks good, a few things to fix/suggestions:

- We need to add Apache license headers to all the new files
- I'm not wild about the -hl:wrapper flag. I think the default should be that users do the wrapping themselves in the mapper creation callback (even for the default mapper).  The reason for this being that you can imagine users extending/modifying the WrapperMapper themselves to add/change functionality. Under such circumstances they will need to add the wrapper themselves and not have the runtime do it for them.  Since the WrapperMapper fully supports the mapper interface, I think this should live completely on the user side (one possible exception to this that I discuss below). It might be good to have a simple example to demonstrate how to do this.
- I'm not super strict about this in most code in the repository, but at least in the runtime code (e.g. runtime.cc) try to stick to an 80 character limit for each line. :)

From an architecture perspective, I find it amusing that we now have a mapper wrapper both on the user side as well as the runtime side (e.g. ConcurrentManager, SerializingManager, ...). Is there anything to be gained by having a unified wrapper? Have we duplicated any functionality on both sides? Is there any information that the runtime has internally that the WrapperMapper could benefit from knowing?

rajshah11:
I have incorporated the changes suggested by you.
I haven't really looked into the other mapper wrappers on the runtime side and am not sure what functionality is provided by them. I will look into it and get back to you.
As far as the WrapperMapper is concerned, the main motivation was to have a debugging framework for the user to print properties set by the runtime and even change some of them during program execution, which is otherwise difficult in a multi-node system like ours. Currently, only select_task_options is wrapped, but I think this could be extended similarly to other functions as well. Hence, any information we feel that user can see or even have the freedom to change can be exposed to the WrapperMapper.

lightsighter:
>> I think this could be extended similarly to other functions as well

Yes, that is my thinking too.  If they do it though, they will need to wrap the mappers themselves in the mapper call back.  We should definitely have an example showing how to do this and explaining how the wrapper works.  Do you think you can be responsible for coming up with the example and documentation?

>> I haven't really looked into the other mapper wrappers on the runtime side and am not sure what functionality is provided by them

Mainly they provide the support for the different mapper synchronization models (e.g. serialized/concurrent/etc). They also provide support for calling back into the runtime from inside mapper calls and doing the necessary pre-emption.  You can imagine there being a way to call from inside a mapper back out to the wrapper interface.

rajshah93:
>> Do you think you can be responsible for coming up with the example and documentation?

When I said that the wrapper mapper can be extended to other mapper calls, I meant that it is something we can do from our side in the future. However, we surely need to have an example showing how to wrap mappers. I can come up with an example and documentation. Can I take an existing example and wrap its mappers or does it have to be a new example altogether?  

>>Is there anything to be gained by having a unified wrapper?

Since the main issue in providing a command-line debugging interface is the concurrency of mapper calls, I think something like SerializingManager will be very useful. The difference being here we have to serialize calls across all mappers.

lightsighter:
I think it's fine to build on an existing example.

Take a look at the internal mapper wrappers and let me know if you have any questions.

Issue #168, opened 2016-09-03
Build error on OpenSuse (i686)


junghans:
It fails with `error: 'LegionRuntime::Arrays::Point<1u>::Point(unsigned int)' cannot be overloaded`:
```
[   61s] cd /home/abuild/rpmbuild/BUILD/legion-legion-16.08.0/build/runtime && /usr/bin/c++   -DCOMPILE_TIME_MIN_LEVEL=LEVEL_DEBUG -DDEBUG_LOW_LEVEL -DGASNETI_BUG1389_WORKAROUND=1 -DGASNETT_THREAD_SAFE -DGASNET_PAR -DLowLevelRuntime_EXPORTS -DREALM_USE_HWLOC -DUSE_GASNET -D_GNU_SOURCE -D_REENTRANT -I/home/abuild/rpmbuild/BUILD/legion-legion-16.08.0/runtime -I/home/abuild/rpmbuild/BUILD/legion-legion-16.08.0/runtime/legion -I/home/abuild/rpmbuild/BUILD/legion-legion-16.08.0/runtime/realm -isystem /usr/include/ibv-conduit  -fomit-frame-pointer -fmessage-length=0 -grecord-gcc-switches -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -g -DNDEBUG -O2 -g -DNDEBUG -fPIC   -o CMakeFiles/LowLevelRuntime.dir/realm/inst_impl.cc.o -c /home/abuild/rpmbuild/BUILD/legion-legion-16.08.0/runtime/realm/inst_impl.cc
[   62s] In file included from /home/abuild/rpmbuild/BUILD/legion-legion-16.08.0/runtime/accessor.h:25:0,
[   62s]                  from /home/abuild/rpmbuild/BUILD/legion-legion-16.08.0/runtime/realm/instance.h:26,
[   62s]                  from /home/abuild/rpmbuild/BUILD/legion-legion-16.08.0/runtime/realm/profiling.h:29,
[   62s]                  from /home/abuild/rpmbuild/BUILD/legion-legion-16.08.0/runtime/realm/faults.cc:19:
[   62s] /home/abuild/rpmbuild/BUILD/legion-legion-16.08.0/runtime/arrays.h:218:18: error: 'LegionRuntime::Arrays::Point<1u>::Point(unsigned int)' cannot be overloaded
[...]

PR #170, opened 2016-09-04
cmake: realm/timers.cc need -fno-strict-aliasing

junghans:
This was discovered in #168.


streichler:
@junghans Actually, can we just set this for all Legion source files?  There's several other places where type-punning goes on, and I'm not sure why g++ only complained about that one file.

junghans:
Done!

junghans:
@chuckatkins please let me know if you have any concerns about this

streichler:
I was just waiting on Travis for this one...  :)

PR #177, opened 2016-09-12
cmake: install headers in legion subdir (fixes #172)

### LINKED ISSUE #172: common.h should not be installing in /include directly
The low-level runtime headers, e.g. `common.h`, get installed in `/include` directly without a subdirectory, that calls from problems.

/cc @tuxfan

junghans:


streichler:
Hrm...  This is much more disruptive than I expected.  Assigning to myself to form a more educated opinion of what needs to happen here.

junghans:
I did it in the least destructive way, but I think it would certainly be nicer to have the code structured inside `/runtime` in the same way as the headers later get exposed to the user.

junghans:
I have a clean, but even more disruptive, version of what I would do here:  junghans/legion@cbd0f6db84175e89de762be7176f099f9cada251

PR #179, opened 2016-09-12
cmake: fix build of tutorials

junghans:
The tutorials got moved to a different folder, but cmake was updated!


junghans:
@chuckatkins : please review!

PR #183, opened 2016-09-20
.travis.yml: added cmake build

junghans:


streichler:
@junghans Thanks for this pull request.  I'm not sure we want to be adding any more work to Travis, as the throughput is pretty poor already, but we definitely want to add cmake coverage to the Gitlab-based CI flow, and we can use this as a template for that.  (@elliottslaughter is the guy who gets to do that.)

elliottslaughter:
@junghans Yes, please head over to [.gitlab-ci.yml](https://github.com/StanfordLegion/legion/blob/master/.gitlab-ci.yml). Dependencies can go in the [Dockerfile](https://github.com/StanfordLegion/legion/blob/master/docker/Dockerfile.gitlab-ci).

When you do, I'd probably restructure things so that cmake replaces make for the entire build. I'd change the environment variable to `TEST_CMAKE` to indicate that the test infrastructure is what cares about cmake (rather than the build itself). Ideally, every test should run under cmake, but we can probably skip Regent at the beginning (there is already a noregent entry for that). Also, it would be ideal if cmake respected the same set of environment variables as make, so that wouldn't need to be changed.

streichler:
@elliottslaughter How hard is it to have some of the gitlab-ci builds use the old make flow and some use cmake?  We can't drop coverage of the old flow for a while at least.

elliottslaughter:
@streichler That's no problem as long as @junghans adds a variable `TEST_CMAKE` like I suggested that controls whether cmake is used or not. Then we can just run some tests with a value of 1 and others with a value of 0.

junghans:
I and @gshipman would really prefer having the cmake tests on github as all of LANL's code are here. Travis supports compiler caching via `ccache` together with cmake and that will decrease the buildtime significantly in the future.

So wouldn't it make more sense to move the _legacy_ make-based buildsystem to gitlab instead.

streichler:
@junghans The Makefile flow isn't legacy yet - it's still the default, so Travis has to test it.  We do want to switch over to cmake "soon", but we're not ready yet, and Travis is already too slow, so we can't add anything to it unless we speed up the existing stuff.  In that vein, the ccache suggestion sounds promising - how big a change would that be to our current (admittedly-fragile) make flow?

junghans:
Have look at 88e6817319fb3440ffe3787695eb1df0dee38781, that should do it for gcc and clang on Linux. Once the build is done, re-trigger it to see if the cache is used.

junghans:
And we could store gasnet and LuaJIT in the cache as well:
```
cache:
  directories:
  - $HOME/build/LuaJIT
 - $HOME/gasnet
```

lightsighter:
Yeah, the trick though is to have the cache pre-populated as part of the Docker image that is built, otherwise the cache is worthless since it will be empty for each Travis run since no state is saved across invocations of the Docker image.

elliottslaughter:
I believe Travis has special support for ccache, which is what makes that work.

The bigger issue is that this just isn't where most of our test time is going. For example, a couple months ago when our Legion Spy tests were timing out, it was because the tests themselves were taking too long. (Which is why we were motivated to go look for an alternative in the first place.)

Would it be possible to add cmake tests for Gitlab? We'll definitely come back and add it for Travis too, but in the mean time I'd like to get Gitlab running cmake. In general, I'd like Gitlab to run a superset of the tests Travis runs.

elliottslaughter:
This has been superseded by 1ee30fceb8624f2864b1bcbe9ce146fa21638213.

PR #185, opened 2016-09-26
cmake: make install include dir configurable

junghans:
Workaround for #172, one can set `-DCMAKE_INSTALL_INCLUDEDIR=include/legion` to chance the install location (and the library interface) to workaround header collisions.


junghans:
@chuckatkins : please review!

chuckatkins:
I see the issue this is referencing but I'm still not sure I get the problem it's trying to solve.  IS the issue at hand conflicts because of sharing an installation directory with other projects?

junghans:
Yes, if you share the include directory with libscrotch, scotch won't compile.
Long term headers like `common.h` should not be installed in the root include, but for now we can put them in a sub directory using `CMAKE_INSTALL_INCLUDEDIR` and can still have the lib directory with other projects.

chuckatkins:
I don't particularly like the idea of the include dir being configurable since it opens the door for a lot of inconsistencies on how legion will be deployed by various people.  I'd rather see the issue addressed in how the headers are organized within the project itself and then still install into a consistent directory.

chuckatkins:
Perhaps to re-org to have:
runtime/lowlevel
runtime/highlevel
runtime/mappers
runtime/realm

And then always install headers from runtime/ into include/legion/.  The only directory to be added to -I then would be /path/to/install/include/legion for the install interface or /path/to/source/runtime for the build interface.  It would also remove the need to have multiple legion include dirs specified in the interface.   It would be a more invasive change but you'd end up with, at least in my opinion, a more maintainable and better code layout.

junghans:
See my comments on #177, this will be a destructive change.

junghans:
And hence I like a band-aid for now, beside the fact that it is confusing to have `CMAKE_INSTALL_INCLUDEDIR` exposed as a configure option, but being functional.

chuckatkins:
I did see that.  from what I could tell though it's still being reviewed and discussed but wasn't rejected.  Can we not use that as the the path forward with this issue being resolved as a side effect?

streichler:
I'm still not liking #177 much, but haven't had a chance to propose an alternative.  Ideally a 'make install' would only copy over "public" header files, and that's also at odds with a "line things up so we can just copy the whole directory tree" approach.

Doesn't the cmake version of 'make install' also create a config file that cmake uses to find things like include files?  Why does the end user have to worry about different choices of `CMAKE_INSTALL_INCLUDEDIR` by the installer?

@junghans Would the issue that this is trying to address go away if you weren't installing Legion into `/usr` or `/usr/local`?

chuckatkins:
> I'm still not liking #177 much, but haven't had a chance to propose an alternative. Ideally a 'make install' would only copy over "public" header files, and that's also at odds with a "line things up so we can just copy the whole directory tree" approach.

I certainly agree on this.  Is there a defined set of "public" headers that I can use to take a shot at this with?

> Doesn't the cmake version of 'make install' also create a config file that cmake uses to find things like include files?

Yes it does but that also assumes that a user can only consume the resulting build with CMake.  It's not a particularly unreasonable assumption to make but if things already need to go into a subdirectory then why not just always put it in the same spot.

> @junghans Would the issue that this is trying to address go away if you weren't installing Legion into /usr or /usr/local?

Not really no.  The issue at hand is sharing an install directory with anything else.  Even if you're not installing to /usr or /usr/local you'll have the same problem if legion is one of may packages that make up a larger project.  This happens to *always* be the case when using /usr or /usr/local but it certainly come up in other situations.

junghans:
@chuckatkins is correct, the problem will show up again in one way or another if you don't use subfolders. 

@streichler: If you tell us, which headers should be installed I can do that, too. However, `common.h`, which originally triggered #172, gets included in multiples place include realm, so it still needs to get installed somewhere.

And to clarify again, this pull request is just a band-aid not a fix for #172. and hence meant as a workaround until you find a version of #177, you like.

elliottslaughter:
Seeing as #172 has been resolved, can we close this?

streichler:
With no votes against, the motion passes.  Closing...

PR #188, opened 2016-09-29
CMake updates

### LINKED ISSUE #187: CMake build breaks on AIO
When I try to build the examples with CMake on default settings, I get a failure to link against Posix AIO. Is there a flag that controls this? (And why wouldn't it auto-detect properly?)

```
$ mkdir build
$ cd build/
$ cmake -DLegion_BUILD_EXAMPLES=ON ..
...
$ make
...
[ 65%] Linking CXX executable ../../bin/dynamic_registration
../../lib/librealm.a(lowlevel_dma.cc.o): In function `LegionRuntime::LowLevel::PosixAIOWrite::launch()':
/home/elliott/legion/runtime/lowlevel_dma.cc:2034: undefined reference to `aio_write'
...
```

Also, more generally, is there documentation for the flags supported by the current CMake build---or could we add some? Does CMake respect environment variables at all, particularly the ones we had set up in the old build system?

Thanks.

CC: @chuckatkins @junghans

chuckatkins:
Adds GASNet conduit preferences and fixes #187

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,15 +73,12 @@ endif()
 #------------------------------------------------------------------------------#
 option(Legion_USE_GASNet "Enable the distributed GASNet backend" OFF)
 if(Legion_USE_GASNet)
-  set(GASNet_THREADING "par" CACHE STRING "GASNet Threading model to use")
-  set_property(CACHE GASNet_THREADING PROPERTY STRINGS seq par parsync)
-  mark_as_advanced(GASNet_THREADING)
+  set(GASNet_THREADING par)
+  set(GASNet_PREFERRED_CONDUITS aries gemini ibv)
+  find_package(GASNet REQUIRED)
   if(NOT GASNet_THREADING STREQUAL "par")
     message(FATAL_ERROR "GASNet threading mode \"${GASNet_THREADING}\" is not currently supported by Legion")
   endif()
-  set(GASNet_CONDUIT "" CACHE STRING "GASNet communication conduit to use")
-  mark_as_advanced(GASNet_CONDUIT)
-  find_package(GASNet REQUIRED)
   install(FILES ${Legion_SOURCE_DIR}/cmake/FindGASNet.cmake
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -73,15 +73,12 @@ endif()
 #------------------------------------------------------------------------------#
 option(Legion_USE_GASNet "Enable the distributed GASNet backend" OFF)
 if(Legion_USE_GASNet)
-  set(GASNet_THREADING "par" CACHE STRING "GASNet Threading model to use")
-  set_property(CACHE GASNet_THREADING PROPERTY STRINGS seq par parsync)
-  mark_as_advanced(GASNet_THREADING)
+  set(GASNet_THREADING par)
+  set(GASNet_PREFERRED_CONDUITS gemeni aries ibv)
```
streichler: This looks like a speeling error on "gemini"?  Also, let's list aries before gemini - I'm not sure if Aries-based systems claim to support gemini or not.

streichler:
@chuckatkins Minor tweak is needed to CMakeLists.txt - see the comment.

CMakeLists.txt
chuckatkins: Good catch on the typo.  Aries and Gemini should be mutually exclusive.  It's still just the default configuration subject to user change.  If you do have both available then chances are likely you'll want to explicitly set the desired conduit anyways.  I don't mind re-ordering since I need to fix the typo anyways but I don't believe it will be an issue that comes up.

streichler:
Hmm...  How hard would it be to not use the cmake 3.6 features?  Just as we're trying not to force everybody to use the most recent versions of compilers, I'd like to avoid forcing people to download and build cmake as part of building Legion.

chuckatkins:
It's easy to do, I just got distracted by other deadlines.  I'll get this
fixed asap and make sure I'm testing with the exact version of CMake that
is specified.

chuckatkins:
Syntax error fixed by #207

PR #189, opened 2016-09-29
CMake: Add version info to project and libraries

### LINKED ISSUE #175: Shared libraries miss an so version
cmake doesn't define an so version for the shared libs.

I will submit a pull-request shortly.

chuckatkins:
Closes #175

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@
 #------------------------------------------------------------------------------#
 
 cmake_minimum_required(VERSION 3.1)
-project(Legion)
+project(Legion VERSION 16.08.0)
 
 #------------------------------------------------------------------------------#
 # Some boilerplate to setup nice output directories
@@ -34,9 +34,6 @@ endif()
 if(NOT CMAKE_RUNTIME_OUTPUT_DIRECTORY)
   set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${Legion_BINARY_DIR}/bin)
 endif()
-
-#library api version, bump from time to time
-set(SOVERSION 1)
 
 #------------------------------------------------------------------------------#
 # More boilerplate to deal with build type and lib type
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -130,7 +130,7 @@ target_link_libraries(LowLevelRuntime
   PRIVATE ${CMAKE_DL_LIBS} ${CMAKE_THREAD_LIBS_INIT}
 )
 set_target_properties(LowLevelRuntime PROPERTIES POSITION_INDEPENDENT_CODE ON)
-set_target_properties(LowLevelRuntime PROPERTIES SOVERSION ${SOVERSION})
+set_target_properties(LowLevelRuntime PROPERTIES VERSION ${Legion_VERSION})
 target_compile_definitions(LowLevelRuntime
   PRIVATE
     $<$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>:DEBUG_LOW_LEVEL>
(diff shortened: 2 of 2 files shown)


chuckatkins:
@junghans are you good with this?

chuckatkins:
Just using one of VERSION or SOVERSION will copy the value to both properties.  If we want to separately track stable APIs with an SOVERSION different than VERSION then we can certainly do that later but at the moment it seems that each release is binary-and API incompatible with the previous.

streichler:
@chuckatkins Using our Legion "version number" in this way is likely to cause folks to assume we're using semantic versioning, which isn't the case.  I'd like to keep the SOVERSION distinct from our release labeling.

chuckatkins:
AFAIK there's no expectation of ABI compatibility right?  If two builds have the same SOVERSION then the expectation is that they are ABI compatible, which is not currently the case from one "release" to the next.  Keeping a distinctly separate SOVERSION that needs to be explicitly changed every release just seemed like unnecessary overhead since the release version is changing anyways, hence why I made it the same as the release version.

Or am I misunderstanding the release process and there is actually a stable ABI to set SOVERSION to?

streichler:
Hmm...  My point is that e.g. versions 16.05.0 and 16.06.0 are not API compatible, which would confuse somebody who thought 16 was the "major" version number.  You're absolutely right though that a constant value of 1 for SOVERSION doesn't really work either, but if there are other tools that expect to see some number there, at least it creates a number we can start consciously incrementing once our API-incompatible changes become less frequent.

chuckatkins:
I don't believe that anything particularly needs it.  @junghans What was the motivation for setting SOVERSION in the first place?  If there's nothing meaningful to set it to then I'd argue for just keeping it empty entirely rather than a constant.

chuckatkins:
With CMake in particuler, VERISON is used for the actual library name and SOVERSION is used for symlinking.  So, in this case, setting the VERSION property to the actual version and SOVERSION to 1 gives us:
```
       14 Sep 29 15:30 liblegion.so -> liblegion.so.1
       19 Sep 29 15:30 liblegion.so.1 -> liblegion.so.16.8.0
 67732584 Sep 29 15:30 liblegion.so.16.8.0
       13 Sep 29 15:30 librealm.so -> librealm.so.1
       18 Sep 29 15:30 librealm.so.1 -> librealm.so.16.8.0
 21958888 Sep 29 15:30 librealm.so.16.8.0
```
with the SONAME being set in the library to `liblegion.so.1` and `librealm.so.1`, which is what will actually get linked to and searched for at runtime.

chuckatkins:
The implications for the different options are

1. Leave both VERSION and SOVERSION empty:
  - `liblegion.so` is the only file created
  - Applications link to `liblegion.so` at compile time.
  - Applications look for `liblegion.so` at runtime
2.  Set VERSION and leave SOVERSION empty (which implicitly sets SOVERSION to VERSION)
  - `liblegion.so.16.8.0` is created with the symlink `liblegion.so -> liblegion.so.16.8.0`
  - Applications link to `liblegion.so` at compile time
  - And applications explicitly look for `liblegion.so.16.8.0` at runtime.
3.  Set VERSION to release version and SONAME to something arbitrary, like `1`
  - `liblegion.so.16.8.0` is created with the symlinks `liblegion.so.1 -> liblegion.so.16.8.0` and `liblegion.so.1 -> liblegion.so.16.8.0`
  - Applications link to `liblegion.so` and compile time.
  - And applications explicitly look for `liblegion.so.1` at runtime.

chuckatkins:
@streichler, @junghans given the various versioning options, any thoughts on how you'd like version information to be expressed in the library files?

streichler:
I think I'd prefer to go back to having VERSION and SOVERSION both empty.  The issue with having to remember to bump the SOVERSION applies to the VERSION as well.

chuckatkins:
Replaced by #191

PR #191, opened 2016-10-03
CMake remove library version info

chuckatkins:
@streichler @junghans Replaces #189 and cancels #175


streichler:
@junghans This PR has been sitting around forever, and I had forgotten all of the context.  It looks like #189 has most of the discussion related to this, and I think we left it at me wanting to remove the SOVERSION stuff, but forgetting to ask you if it was necessary for any of your needs and/or the packages being made.  Do you think removing it would be ok?  (If you're unsure, we'll probably leave it in, since it's not breaking anything right now.)

streichler:
Closing this as the current version seems adequate for current needs.

PR #196, opened 2016-10-19
cmake: use $MPI_INCLUDE as hint for gasnet

junghans:
Using $MPI_INCLUDE as a hint will help cmake to find mpi-enabled version of gasnet in Fedora.

/cc: @tuxfan

### DIFF
--- a/cmake/FindGASNet.cmake
+++ b/cmake/FindGASNet.cmake
@@ -185,6 +185,8 @@ if(NOT GASNet_FOUND AND NOT TARGET GASNet::GASNet)
   mark_as_advanced(GASNet_ROOT_DIR)
   if(GASNet_ROOT_DIR)
     set(_GASNet_FIND_INCLUDE_OPTS PATHS ${GASNet_ROOT_DIR}/include NO_DEFAULT_PATH)
+  else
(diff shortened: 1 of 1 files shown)


streichler:
I have no objections, as long as it doesn't break if `MPI_INCLUDE` is not present in the environment.

@chuckatkins , your thoughts?

chuckatkins:
While I'm not opposed to the idea, you would want to implement it a bit differently, like this:
```cmake
if(GASNet_ROOT_DIR)
  set(_GASNet_FIND_INCLUDE_OPTS PATHS ${GASNet_ROOT_DIR}/include NO_DEFAULT_PATH)
else()
  set(_GASNet_FIND_INCLUDE_OPTS HINTS ENV MPI_INCLUDE)
endif()
find_path(GASNet_INCLUDE_DIR gasnet.h ${_GASNet_FIND_INCLUDE_OPTS})
```
This ensures that if GASNet_ROOT_DIR is specified then it's used exclusively.  Also the "ENV VARNAME" syntax rather than $ENV{VARNAME} is specific to find_* commands and takes some additional steps to deal with some OS level escaping issues.

junghans:
Done!

PR #197, opened 2016-10-20
Configfile

zard49:
Add support for a Lua based configuration file for setting runtime options rather than specifying them on the command line.

Default config file is "config.lua" or it can be specified with -ll:lua_config

Lua variables corresponding to runtime variables start with the prefix “opt_” so –level becomes opt_level. “:” is replaced by “_” so “–ll:cpu” becomes opt_ll_cpu etc.

This allows for example to only specify an option on a given node e.g.

if hostname() == "foo" then
  opt_level = "realm=0"
end

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,17 @@ if(Legion_USE_HWLOC)
 endif()
 
 #------------------------------------------------------------------------------#
+# LuaJIT configuration
+#------------------------------------------------------------------------------#
+option(Legion_USE_LuaJIT "Use LuaJIT for config file" OFF)
+if(Legion_USE_LuaJIT)
+  find_package(LuaJIT REQUIRED)
+  install(FILES ${Legion_SOURCE_DIR}/cmake/FindLuaJIT.cmake
+    DESTINATION lib/cmake/Legion
+  )
+endif()
+
+#------------------------------------------------------------------------------#
 # GASNet configuration
 #------------------------------------------------------------------------------#
 option(Legion_USE_GASNet "Enable the distributed GASNet backend" OFF)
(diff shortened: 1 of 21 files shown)


elliottslaughter:
Maybe we could tweak this to follow Lua's traditional approach to modules (where the script returns the object as a "return value"). That way you wouldn't need a prefix on the option names, and you wouldn't be stuffing things into the global environment.

```
local options = {}
options.ll_cpu = 4
if hostname() == "foo" then
  options.level = "realm=0"
end
return options
```

I'm not set on doing it this way, but it does seem like a slightly cleaner approach to me. Thoughts?

streichler:
I've been spending a few cycles thinking about this for a while now, and there's some other features I'd like to get in as well.  In particular, I'd like to make sure that we catch typos in config variable names and that the Lua script have some information available to it with respect to defaults/etc.  The basic idea is similar - to turn the CommandLineParser into a smarter object that can get its data from several places, so I can probably reuse a bunch of this code.

Then there's also the matter of tying this in to those Lua processors that I'll be adding "any day now".

zard49:
Sean,

Let me know if there are things I can help on. I made this is as proof of concept based on our talking the last time you were out here. I figured I'd make a pull request just to the get the ball started rolling. 

-dp

streichler:
@zard49 Just letting you know I haven't forgotten this, but it's not yet up to the top of my list.

zard49:
I've been messing around w/ a possible idea for catching typos, using lua metatables. I'll let you know if I get somewhere... Basically CommandLineParser.add_option_* would add entries to the lua option table, and then at the parse stage we would use the lua metatable __newindex stuff to set this table so we would get an error if a new key(which would be a typo of an existing option) was attempted to be added to the option table.

streichler:
No updates on this in the last year, but still something I'd like to work on, so I'm leaving this open and assigned to me.

PR #198, opened 2016-10-21
Bergen

tuxfan:
This fixes a bug in cmake/FindGASNet.cmake with an improperly formatted 'endif' statement.

PR #199, opened 2016-10-21
cmake: syntax error

junghans:
introduced in a76904852c60b5b2d80508e32cd69694c2aff1fc, #196


streichler:
@junghans can you compare to what @tuxfan has in #198?  He's got an additional test for the existence of `MPI_INCLUDE` that seems like it might be useful.

junghans:
Yeah, just saw that. I think the check is not necessary as  `_GASNet_FIND_INCLUDE_OPTS` is only use as a hints, but let's ask @chuckatkins .

PR #205, opened 2016-11-04
cmake: link mpi when using mpi conduit of gasnet

junghans:
gasnet uses static libraries by default, which have no linking
information. Using the mpi conduit will lead to undefined symbols
somewhere down the line, so let's add it to the link libraries.

/cc @tuxfan @chuckatkins


chuckatkins:
You shouldn't need to link them explicitly.  While the mpi libs won't get
pulled in from the gasnet library, the find module doesn't actually look
for any libraries.  Instead it parses the usage information from the
conduit's make file (which is why this module is so complicated w.r.t. a
typical find module), which should be including them.  If they're not
coming through then something is missing from the make file parsing and
that's where it should be addressed rather than calling find_package(MPI).

junghans:
The mpi library is missing due to the fact that in e.g. `/usr/include/mpi-conduit/mpi-par.mak` `mpicc` is used for linking, which will pull `libmpi` automatically.

We could of course force legion to be linked with `mpicc` as well.

chuckatkins:
Ugh, gasnet's build is a mess.  We shouldn't force mpicc but if we can tack
on the appropriate verbose flag into the temporary makefile used for
parsing that should expand the necessary underlying compiler flags to work
with the existing parsing.

junghans:
This is really hard, not all mpi compilers are called `mpi*` nor all of them support `mpicc --showme`. On the distribution stack side, I worked around that issue by making shared libs:
http://pkgs.fedoraproject.org/cgit/rpms/gasnet.git/tree/gasnet.spec#n131 

Btw, here is where we found the issue:
https://travis-ci.org/losalamos/flecsi-third-party/builds/173362341

chuckatkins:
I'm pretty sure there is a much simpler way to do it with what cmake
already knows about compilers.  I'll dig through some docs and get a pr up
tomorrow.

chuckatkins:
So, my grand plan didn't quite work and I still had to fall back to find_package(MPI), i just limited the condition a bit more.  See #206

junghans:
Replace by #206.

PR #216, opened 2017-01-19
Cmake from lanl2

ipdemes:
A fix for the CMake build issue with propagation compile-time definitions to other projects, that are built on top of the Legion:
1)  CMAKE generates legion_defines.h file where it stores all definitions
2) In case Legion is built with CMake, legion_defines.h file gets included into legion.h

@chuckatkins, @streichler : please review

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ set(SOVERSION 1)
 set(CMAKE_BUILD_TYPE  Debug CACHE STRING "Choose the type of build" FORCE)
 set(BUILD_SHARED_LIBS OFF   CACHE BOOL   "Whether or not to build shared libraries instead of static")
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DLEGION_CMAKE")
+add_definitions(-DLEGION_CMAKE)
 
 #https://github.com/StanfordLegion/legion/issues/168#issuecomment-244582958
 include(CheckCXXCompilerFlag)
@@ -54,7 +54,7 @@ if(COMPILER_SUPPORTS_NO_STRICT_ALIASING)
 endif()
 
 if (${CMAKE_BUILD_TYPE} EQUAL "Debug")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DDEBUG_REALM -DDEBUG_LEGION")
+  add_definitions(-DDEBUG_REALM -DDEBUG_LEGION)
   set(DEBUG_REALM ON)
   set(DEBUG_LEGION ON)
 endif()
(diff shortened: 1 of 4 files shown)


streichler:
Looks ok.  I'll press the button as soon as the Travis build finishes.  Thanks!

ipdemes:
@streichler : thanks!

chuckatkins:
What problem is this trying to solve? The compile time definitions should already be getting propagated via the usage requirements.  If they're not then we should fix it there.  Explicitly adding it to the CMAKE_<LANG>_COMPILE_FLAGS will short circuit the usage requirements and actually prevent them propagating correctly.

ipdemes:
@chuckatkins : I wasn't sure about  CMAKE_CXX_FLAGS so your feedback would be very helpful. I can replace them with add_definitions, when necessary.
The main purpose of this is generating legion_defines.h file that includes all Legion's compile-time definitions.

chuckatkins:
The -D flags your adding are already being used to compile Legion.  They're not currently propagated because they're seen as private rather than public definitions, meaning they're only needed to build legion, but not by anything that uses legion.  For instance, Legion needs to know if it's built with GASNet but a Legion based application doesn't so USE_GASNET is a private compile definition that doesn't propagate, rather than a public one.  However, the debug logging level is used in the public headers so COMPILE_TIME_MIN_LEVEL is a public compile definition that will propagate.

Are there definitions that are currently missing or that should be propagated as public that aren't because they're currently private?

ipdemes:
@chuckatkins : Are these definitions propagated through *.cmake files?
In FleCSI we treat Legion as a stand-along library: we link to liblegion.so, librealm.so and include directory where it was installed (so we don't use *.cmake files here). In this way FLeCSI doesn't know anything about definitions Legion was build with. This results in having very interesting errors when we include "legion.h" in our code. 
So far we need following definitions:
#define DEBUG_REALM 
#define DEBUG_LEGION
#define PRIVILEGE_CHECKS
#define BOUNDS_CHECKS
#define GASNET_CONDUIT_MPI/UDP/IBV ..
#define GASNETI_BUG1389_WORKAROUND 1

ipdemes:
@chuckatkins : would you like it more with the next change b9a6e580088cef77ba381336745db98f80ef5eed ?

CMakeLists.txt
```diff
@@ -44,13 +44,21 @@ set(SOVERSION 1)
...
+
 #https://github.com/StanfordLegion/legion/issues/168#issuecomment-244582958
 include(CheckCXXCompilerFlag)
 check_cxx_compiler_flag("-fno-strict-aliasing" COMPILER_SUPPORTS_NO_STRICT_ALIASING)
 if(COMPILER_SUPPORTS_NO_STRICT_ALIASING)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-strict-aliasing")
 endif()
 
+if (${CMAKE_BUILD_TYPE} EQUAL "Debug")
+  add_definitions(-DDEBUG_REALM -DDEBUG_LEGION)
```
junghans: Isn't that unneeded as `DEBUG_REALM` is part of `legion_defines.h`, which gets included?

ipdemes: @junghans:  legion_defines.h gets generated after:
add_definitions(-DDEBUG_REALM -DDEBUG_LEGION)

I guess instead of "add_definitions" I should do "set (DEBUG_REALM ON)" ?

junghans: And the build, which actually uses the defines, happen after generating `legion_defines.h`.

ipdemes: @junghans : true. This commit should help with avoiding redefinitions: 097746c8941b101b3827dc67d21f2e31fc25e783

Issue #181, opened 2016-09-20
Compilation failure on Titan when using gasnet


sdalton1:
Compilation fails on Titan when USE_GASNET is enabled. 

Stripped down command that fails
```shell
CC -o /ccs/home/sdalton/legion/runtime/legion/legion_views.o -c /ccs/home/sdalton/legion/runtime/legion/legion_views.cc  -DGASNETI_BUG1389_WORKAROUND=1 -DUSE_GASNET -I/ccs/home/sdalton/legion/runtime -I/ccs/home/sdalton/legion/runtime/realm -I/ccs/home/sdalton/legion/runtime/legion -I/ccs/home/sdalton/legion/runtime/mappers -I/ccs/home/sdalton/gasnet/release/include -I/ccs/home/sdalton/gasnet/release/include/gemini-conduit
```

Debug session with Mike revealed that one of the gasnet headers is defining a conflicting macro before the legion_utilities.h file is included.

```c++
#define __STDC_FORMAT_MACROS
#include <inttypes.h>
#define MASK_FMT "%16.16" PRIx64
```

```shell
/ccs/home/sdalton/legion/runtime/legion/legion_utilities.h: In member function 'char* Legion::BitMask<T, MAX, SHIFT, MASK>::to_string() const':
/ccs/home/sdalton/legion/runtime/legion/legion_utilities.h:49:27: error: expected ')' before 'PRIx64'
 #define MASK_FMT "%16.16" PRIx64
```

lightsighter:
The problem is GASNet is including inttypes.h before this one in the include hierarchy for legion_utilities.h, but they are not prefixing it with the __STDC_FORMAT_MACROS define.  As a result the PRIx64 macro is not being defined.

streichler:
@sdalton1 can you try adding -D__STDC_FORMAT_MACROS to your CC_FLAGS?  If that works, we'll jam it into the standard CC_FLAGS - we've seen one other system that needs this too.

sdalton1:
Adding __STDC_FORMAT_MACROS fixes the issue. The only side effect is a warning about redefining __STDC_FORMAT_MACROS.

```shell
/ccs/home/sdalton/legion/runtime/legion/legion_utilities.h:48:0: warning: "__STDC_FORMAT_MACROS" redefined
 #define __STDC_FORMAT_MACROS
```

streichler:
@sdalton1 This should hopefully be fixed (sans warning) with 208a21d.

[...]

PR #217, opened 2017-01-27
Added "-D__STDC_FORMAT_MACROS" flag to the cmake build system. Appare…

ipdemes:
…ntly we need this flag to get Legion built on some systems

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,6 +148,10 @@ add_subdirectory(runtime)
 # configure header
 #------------------------------------------------------------------------------#
 # Checking for all defines in the CXX Flags
+
+add_definitions(-D__STDC_FORMAT_MACROS)
+set (__STDC_FORMAT_MACROS ON)
+
 string(REPLACE " " ";" FLAGS_LIST ${CMAKE_CXX_FLAGS})
 FOREACH(FLAG ${FLAGS_LIST})
   string (FIND ${FLAG} "-D" START_STR)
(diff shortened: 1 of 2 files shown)


ipdemes:
This is related to the issue:
https://github.com/StanfordLegion/legion/issues/181

streichler:
@ipdemes : I don't think this will work.  On the systems with this issue, this define must be set before
the first #include of inttypes.h, which may come before any legion header is parsed in an application
that includes other headers.  I'm pretty sure it needs to be on the compilation command line for
application source files to work in all cases.

junghans:
I think legion should define `__STDC_FORMAT_MACROS` in the headers, which use `inttype.h` and friends.

streichler:
@junghans That's what we started with.  It works as long as the application code that is including legion.h hasn't already included inttypes.h (directly or indirectly).  The `__STDC_FORMAT_MACROS` macro must be defined before the first include of inttypes.h, no matter where it is, and a command line parameter is the only way to guarantee that.

junghans:
You could error out if ` _INTTYPES_H` is defined without `__STDC_FORMAT_MACROS` before including `inttypes.h` in `runtime/legion/legion_utilities.h`. Something like:
```
#ifndef __STDC_FORMAT_MACROS
#ifdef  _INTTYPES_H
#error Legion need to be build with -D__STDC_FORMAT_MACROS
#endif
#else
#define __STDC_FORMAT_MACROS
#endif
#include <inttypes.h>
```

Plus, with all these defines legion really needs a pkg-config file.

ipdemes:
@streichler, @junghans  : this commit actually fixes an error we see without __STDC_FORMAT_MACROS on one of our clusters (Darwin). I think this works because legion_defines.h gets included before any other legion header file is included.  I have just double-checked it this morning.

streichler:
I'll merge this for now, but it sounds like you're just getting lucky.  I'll look into fixing it "right"
(i.e. eliminating the need for this define in application code at all, unless it also uses the format
macros).

ipdemes:
@streichler : thank you. Do you have any test/example in legion that might  break? I can try  to check it out with cmake.

PR #229, opened 2017-03-13
cmake: enable testing

junghans:
```
cmake -DLegion_ENABLE_TESTING=ON .
make
make test
```


junghans:
@chuckatkins

junghans:
@elliottslaughter: I patched that into the fedora package, so it would be nice to have something like that as a part in the next release.

PR #244, opened 2017-05-01
Updated FindGASNET function to add MPI include path

tuxfan:
The mpi_interop example includes mpi.h and needs the MPI include path found in FindGASNet. @junghans


junghans:
Needed on Fedora!

PR #248, opened 2017-05-06
Runtime error documentation

aheirich:
class to let us document legion runtime errors in the source code.  usage is like this:

            LegionRuntime::Logger::MessageDescriptor
            ILLEGAL_REQ_EMPTY_FUTURE(1000,
                                     "this is an html description of what to do if you see the subsequent error");
            Internal::log_run.error(ILLEGAL_REQ_EMPTY_FUTURE.id(), "Illegal request for future "
                                    "value from empty future");
A separate script parses the source files to collate these messages for posting online.
This introduces one new typedef at the Realm level for LoggerMessageID (int).  This might change to a string some day or a long int.
This introduces a new class at the LegionRuntime::Logger level, MessageDescriptor.


lightsighter:
A couple of things:

1. We need to get all the continuous integration tests to pass before doing the merge
https://travis-ci.org/StanfordLegion/legion/builds/229382049?utm_source=github_status&utm_medium=notification
2. The LegionRuntime namespace is deprecated.  I think we probably want to put this in the Legion::Internal namespace, e.g.
https://github.com/StanfordLegion/legion/blob/master/runtime/legion/runtime.h#L29-L30
3. You were going to define an operator conversion to Realm::LoggerMessageID right?

aheirich:
Mike,
1&2 yes
3 I made a type so the compiler would type check for us and catch errors.  For this reason I did not define an implicit conversion to int for Realm::LoggerMessageID.  Are you ok with this?

lightsighter:
What kind of type errors are you hoping to catch?  I would think that we would always want to be able to pass this to a Realm logging call that takes an integer as the first argument.  It won't implicitly cast to a 'const char*' so it should catch those cases for us.

aheirich:
The realm logging call takes a LoggerMessageID rather than an int precisely so you won't call it unintentionally with an int argument.  The recommended way to call it is with a MessageDescriptor::id().  If you don't call it with a MessageDescriptor your error message will not get collected by the collation mechanism.  If you really want to call it with an int you can still do so by using a cast but I would not want this to happen silently (and I cannot think of a reason to do this).

streichler:
Hmm...  I'm not wild about adding another use of `Legion::...` inside of Realm, but I'm willing to clean that up later.  Two more pressing questions:
1) @aheirich Does the message extractor script fail loudly if the same message ID is used in two places?
2) @lightsighter How painful is this going to be to merge into deppart/control_replication branches on your side?

lightsighter:
I'm pretty concerned about the merge with deppart currently.  I thought the plan was we were going to create the infrastructure, and modify a few of the error messages to test it out, and then wait for the deppart merge before annotating all the error messages.  However, it looks like Alan went through and annotated all the error messages which is going to cause lots of conflicts with deppart, especially in the legion_context.cc and region_tree.cc files. Merging with control replication won't be so bad since that is mostly new code on top of deppart, but deppart is going to be painful.

streichler:
Following up on my question from above, @aheirich does the message extractor script also flag error/fatal messages that do not have a message ID associated with them?  If so, we'd be able to resolve the deppart merge conflicts by ignoring the message ID changes and then add them back in manually without fear of missing any.

aheirich:
I can make the python script notify us of messages that don't have annotations.  But I don't think there should be any, I annotated them all.

aheirich:
to answer @streichler yes the python script will fail if there are duplicate message ids in use
it's not graceful but seems adequate

aheirich:
FYI to all if there is manual work to do I will do it
I already spent the time putting the annotations in!

aheirich:
TO DO (note to self): modify python script to flag error or fatal that are missing annotations.

lightsighter:
@aheirich The two files which I foresee being problematic with the merge with deppart are region_tree.cc and legion_context.cc.  They are the ones we'll probably want to set back to their state with master before the merge and then re-annotate them in deppart after merging master into deppart.  You'll see legion_context.cc has changed a lot, and a large fraction of region_tree.cc was completely rewritten with some of it being refactored into region_tree.inl.

aheirich:
Hi

How do I change my registered email address for this list?

thanks

Alan

elliottslaughter:
You should be able to choose your primary email at https://github.com/settings/emails

aheirich:
Mike,

Is this ready for me to work on yet?  Has deppart merged into master?  Or should I merge deppart into runtime_error_messages?

Alan

PR #254, opened 2017-05-19
realm: allow uthread default override

noahdesu:
The JVM does not take kindly to non-native threads.

Signed-off-by: Noah Watkins <noahwatkins@gmail.com>


streichler:
Thanks Noah!  This pull request reminded me that I needed to add the command-line version as well, which is done in 2a89653.

noahdesu:
Oh, I didn't notice that it was runtime configurable. That is even better. Thanks!

streichler:
It wasn't until five minutes ago.  :)

PR #258, opened 2017-05-23
Api documentation

aheirich:
filled in some missing api documentation
I was unclear about the distinction between Color and DomainPoint in e.g.

       IndexPartition get_index_partition(Context ctx, IndexSpace parent, 
                                          Color color);
+        /**
+         * Return the index partitioning of an index space
+         * at a specified point.
+         * @param ctx enclosing task context
+         * @param parent index space
+         * @param point in index partition
+         * @return handle for the index partition at point
+         */
       IndexPartition get_index_partition(Context ctx, IndexSpace parent,
-                                         const DomainPoint &color);
+                                         const DomainPoint &point);

### DIFF
--- a/runtime/legion/legion.h
+++ b/runtime/legion/legion.h
@@ -3200,12 +3200,34 @@ namespace Legion {
        */
       IndexPartition get_index_partition(Context ctx, IndexSpace parent, 
                                          Color color);
+        /**
+         * Return the index partitioning of an index space
+         * at a specified point.
+         * @param ctx enclosing task context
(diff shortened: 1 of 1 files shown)


lightsighter:
I think I'd rather do this with function grouping.  
https://www.stack.nl/~dimitri/doxygen/manual/grouping.html
We also need to do it for all the context free methods, not just the get_index_partition methods.

lightsighter:
I addressed this issue with this commit instead: 3460272

PR #260, opened 2017-06-06
Cmake defines

ipdemes:
Added "ENABLE_LEGION_TLS" to the legion_defines

### DIFF
--- a/cmake/legion_defines.h.in
+++ b/cmake/legion_defines.h.in
@@ -32,6 +32,9 @@
 #ifndef USE_CUDA
 #cmakedefine USE_CUDA
 #endif 
+#ifndef ENABLE_LEGION_TLS
+#cmakedefine ENABLE_LEGION_TLS
+#endif
 #ifndef GASNETI_BUG1389_WORKAROUND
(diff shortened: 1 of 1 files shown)


lightsighter:
@ipdemes The ENABLE_LEGION_TLS feature is still experimental until we've done some more benchmarking on the performance overhead that it incurs.  I'll keep this pull request around until we decide whether we want to make it part of the mainline Legion master branch or not.

ipdemes:
@lightsighter , I think we need to have this logic in legion_defines as long as you have "#ifdef ENABLE_LEGION_TLS" anywhere in the source code. Otherwise cmake would be broken for those who need this option to be ON. And you can always revert this commit in case you decide to remove LEGION_TLS from the legion src.

lightsighter:
Yes, @streichler has informed me of my ignorance of cmake. :)  Merging the pull request.

ipdemes:
@lightsighter , Thank you!!

elliottslaughter:
@ipdemes If you're done with the branch, can you delete it? (There is button in this thread, or you can run `git push origin --delete cmake_defines`.)

ipdemes:
@elliottslaughter , Sure. It should be removed now

PR #263, opened 2017-06-11
long running test to check for memory leaks or instability

aheirich:
this program running on MacOS 10.12.5 (Sierra) is leaking and cannot complete the index task launch

### DIFF
--- a/test.py
+++ b/test.py
@@ -64,6 +64,10 @@ if platform.system() != 'Darwin':
         ['test/attach_file_mini/attach_file_mini', []],
     ]
 
+legion_cxx_tests = legion_cxx_tests + [
+        ['test/long_running/long_running', []],
+]
+
 legion_gasnet_cxx_tests = [
     # Examples
     ['examples/mpi_interop/mpi_interop', []],
--- a/test/long_running/CMakeLists.txt
+++ b/test/long_running/CMakeLists.txt
@@ -15,14 +15,15 @@
 #------------------------------------------------------------------------------#
 
 cmake_minimum_required(VERSION 3.1)
+project(LegionTest_long_running)
 
+# Only search if were building stand-alone and not as part of Legion
 if(NOT Legion_SOURCE_DIR)
   find_package(Legion REQUIRED)
 endif()
 
-add_subdirectory(attach_file_mini)
-add_subdirectory(long_running)
-
-if(Legion_USE_HDF5)
-  add_subdirectory(hdf_attach_subregion_parallel)
+add_executable(long_running long_running.cc long_running.h)
+target_link_libraries(long_running Legion::Legion)
+if(Legion_ENABLE_TESTING)
+  add_test(NAME long_running COMMAND $<TARGET_FILE:long_running>) 
 endif()
--- a/test/long_running/Makefile
+++ b/test/long_running/Makefile
@@ -1,0 +1,49 @@
+# Copyright 2017 Stanford University
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+ifndef LG_RT_DIR
+$(error LG_RT_DIR variable is not defined, aborting build)
+endif
+
+# Flags for directing the runtime makefile what to include
+DEBUG           ?= 1		# Include debugging symbols
+OUTPUT_LEVEL    ?= LEVEL_DEBUG	# Compile time logging level
+USE_CUDA        ?= 0		# Include CUDA support (requires CUDA)
+USE_GASNET      ?= 0		# Include GASNet support (requires GASNet)
+USE_HDF         ?= 0		# Include HDF5 support (requires HDF5)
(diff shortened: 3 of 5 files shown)


aheirich:
It is wrong to call these memory leaks, they are not showing as leaks under the memory leak tool in XCode instruments.
This is probably the result of policy about when to garbage collect.  This test is demonstrating that memory is not being cleaned up after an interior task exits.

aheirich:
something goes wrong in the tests running on travis, and I don't see anything wrong with the application
the messages below show the interior task is calling the constructor, then exiting

...
fragment domain [(0,0,0),(19,239,3)] num points 19200
executed destructor
fragment domain [(0,0,0),(19,239,3)] num points 19200
executed destructor
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated

aheirich:
@elliottslaughter I think this is an example of a test that we don't want to run on every pull request (because it takes too long) but do want to run regularly, e.g. nightly

aheirich:
here is a different way of dying on an ubuntu system

https://travis-ci.org/StanfordLegion/legion/jobs/241638797

fragment domain [(0,0,0),(19,239,3)] num points 19200
executed destructor
############################################################
### Exiting Stage: legion_cxx
###   * Exception Type: <class 'subprocess.CalledProcessError'>
###   * Elapsed Time: 0:04:44.440301
############################################################
Traceback (most recent call last):
  File "./test.py", line 690, in <module>
    driver()
  File "./test.py", line 687, in driver
    run_tests(**vars(args))
  File "./test.py", line 606, in run_tests
    run_test_legion_cxx(launcher, root_dir, tmp_dir, bin_dir, env, thread_count)
  File "./test.py", line 144, in run_test_legion_cxx
    run_cxx(legion_cxx_tests, flags, launcher, root_dir, bin_dir, env, thread_count)
  File "./test.py", line 134, in run_cxx
    cmd(launcher + [test_path] + flags + test_flags, env=env, cwd=test_dir)
  File "./test.py", line 121, in cmd
    return subprocess.check_call(command, env=env, cwd=cwd)
  File "/opt/python/2.7.12/lib/python2.7/subprocess.py", line 541, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/home/travis/build/StanfordLegion/legion/test/long_running/long_running', '-logfile', 'out_%.log']' returned non-zero exit status -9

lightsighter:
@aheirich This works fine for me as long as I don't try to make partitions with 19.2K subregions per partition which I don't think is what you were intending to do.  
diff --git a/test/long_running/long_running.cc b/test/long_running/long_running.cc
index b3edf55..0827de1 100644
--- a/test/long_running/long_running.cc
+++ b/test/long_running/long_running.cc
@@ -65,9 +65,9 @@ namespace Legion {
       // logical partition
       
       Point<3> fragmentSize;
-      fragmentSize.x[0] = 16;
-      fragmentSize.x[1] = 1;
-      fragmentSize.x[2] = 1;
+      fragmentSize.x[0] = 20;
+      fragmentSize.x[1] = 240;
+      fragmentSize.x[2] = 4;
       Blockify<image_region_dimensions> coloring(fragmentSize);
       imageFragmentIndexPartition = mRuntime->create_index_partition(mContext, region.get_index_space(), colori
       mRuntime->attach_name(imageFragmentIndexPartition, "image fragment index");

aheirich:
Yes, but, but ….

You can argue that this is an unreasonable number of index space points for a 4-node system and I would agree.  But I don’t think the runtime should deadlock or crash in this case, and I even weakly assert that the launch should still complete, albeit with poor performance.  Or die due to out-of-memory.

But on a 10,000 node system this number of points won’t be unreasonable, and won’t even be an upper limit.

There is another problem that this long_running test is showing which is causing me real issues today in my unit tests for visualization reductions.  Please look at the memory usage at it loops on 10,000 calls to interior_task.  The memory never gets reclaimed.  This is causing my unit tests to die.  It looks like garbage collection never gets invoked.

lightsighter:
It's not deadlocking or crashing (unless you run it long enough to run out of memory), it's just takes a very very long time to make 19.2K index subspaces.  All that meta-data can be very expensive.  Furthermore, there is a known leak where the runtime won't collect some meta-data for region tree data structures until the runtime is being shutdown.  The reason for this is that it is very expensive to determine when some of this meta-data can be collected in a distributed system.  Most Legion applications so far make a few region trees and use them repeatedly throughout the lifetime of the application.  Until we have some compelling real applications (not synthetic benchmarks) that actually require the need to dynamically create and destroy logical regions then I don't have any plans to modify the current implementation.

lightsighter:
Also, this isn't how we're going to make 19.2K subspaces with control replication.  This approach is inherently not scalable, but the one implemented in the control replication is, so I'm not worried about making this approach scale.

aheirich:
Ok, understood.  I can rewrite the unit tests to reuse the same logical region if the resolution has not changed.  And I can keep the index spaces small.

lightsighter:
Actually you want to keep the index spaces large and the number of index spaces small.  

In general you should think about making logical regions as being as expensive as creating instances of monads in functional programming languages, it's not something that should be done every function (task) invocation.

streichler:
@aheirich Where do we stand on this?  Is it part of the perf testing stuff you've been working on?

aheirich:
This was a wish list item from me, thinking long term
I recall that Mike was not persuaded this would be valuable at the present time
We can close this

On Oct 10, 2017, at 8:21 AM, streichler <notifications@github.com<mailto:notifications@github.com>> wrote:


@aheirich<https://github.com/aheirich> Where do we stand on this? Is it part of the perf testing stuff you've been working on?

streichler:
Ok, closed.

PR #268, opened 2017-06-16
convert all uses of MessageDescriptor to one of REPORT_LEGION_ERROR, …

aheirich:
…REPORT_LEGION_WARNING, REPORT_LEGION_FATAL


streichler:
I have a minor concern with anything that reinforces the old printf-style debugging messages, but no major concerns.

aheirich:
we also allow multiple messages to use the same id now.  these will be grouped together in the collated message html pages.

lightsighter:
@aheirich You can delete the other deppart branches now once you're done with them

PR #270, opened 2017-06-19
removed wrong semantic for Debug mode

ipdemes:
without this commit DEBUG_LEGION will always be defined.

PR #271, opened 2017-06-21
cmake: allow non-Debug cmake build type

junghans:
Fix #270.

/cc @ipdemes  @elliottslaughter


ipdemes:
@junghans, @elliottslaughter,  thank you!

PR #272, opened 2017-06-21
cmake: allow CMAKE_BUILD_TYPE to be empty

junghans:
Related to #271.

/cc @ipdemes @elliottslaughter


elliottslaughter:
I see how this avoids breaking the build, but it doesn't respect default values of `CMAKE_BUILD_TYPE`. I.e. when the value is not set on the command line, CMake must choose a default build type somehow. Is there no way to just get the definitive value of the build type?

elliottslaughter:
Also, what is the purpose of setting these variables here? As best I can tell, they are not actually being used anywhere, because the conditional inside `runtime/CMakeLists.txt` is on `CONFIG:Debug` not on `DEBUG_LEGION`. This seems undesirable: it's possible for it to get out of sync with the condition that sets DEBUG_LEGION later. Maybe instead of `CONFIG:Debug` we should be checking `DEBUG_LEGION`?

```
$ rg DEBUG_LEGION CMakeLists.txt 
58:  set(DEBUG_LEGION ON)
$ rg DEBUG_LEGION runtime/CMakeLists.txt 
245:    $<$<CONFIG:Debug>:DEBUG_LEGION>
$ rg DEBUG_LEGION cmake/
cmake/legion_defines.h.in
8:#ifndef DEBUG_LEGION
9:#cmakedefine DEBUG_LEGION
```

junghans:
Well, the cmake variable is used in `legion_defines.h.in`, `#cmakedefine` will become a define if the variable is set.

junghans:
I think we should do something like:
```
if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CXX_FLAGS)
  #release comes with -O3 by default
  set(CMAKE_BUILD_TYPE Debug CACHE STRING "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel." FORCE)
endif(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CXX_FLAGS)
```

junghans:
ping @elliottslaughter

elliottslaughter:
@junghans With the latest patch it seems to be having trouble finding legion_defines.h:

https://travis-ci.org/StanfordLegion/legion/jobs/245852498

junghans:
I removed the last commit, that should fix it!

junghans:
@elliottslaughter : Can this be merged?

PR #275, opened 2017-06-28
Build fixes

stelleg:
I had to make these changes to build with cmake 3.8.2 and gcc 7.1.1

The CMakeLists change is a reverting of a previous change. I'm not sure why the change was made but it failed for my version of cmake. I believe `EQUAL` is the wrong comparator for strings. It should be `STREQUAL`. 

I suppose the header shadowing issue could be fixed by fiddling with include ordering, but this works for me.


lightsighter:
@stelleg Does this need to go into the deppart branch or can this go in the master branch?

stelleg:
@lightsighter deppart, I'm not having these issues in master or stable

streichler:
@stelleg Even if it's not strictly needed in master, I'd prefer if it went there.  Can you try it out on the master branch and make sure it works ok?  Thanks.

stelleg:
@streichler it wouldn't do anything on master. One file changed only exists in deppart and the other file already is fixed. 

Should we not be opening issues/PR's for deppart?

streichler:
Note to self: Drink coffee, then try to think.  :)

Yeah, the interval_tree thing definitely should be fixed in deppart, but we haven't (intentionally) made any cmake changes in deppart, so I'm surprised there's a difference there and will go hunting for its source.

More generally, issues/PRs for deppart are welcome, but given that it won't fold into master for a few more weeks at least (and not into stable for a few months), anything that can be fixed upstream in master (or in some cases the dma branch) should be.

streichler:
Sorry for the delay on this, but I was able to determine that the cmake issue was just because we were about a week behind on the master->dma->deppart merges.  I pulled the corresponding master changes through and I believe that removes the need for any cmake-changes specific to deppart.

For the realm header file change, I just cherry-picked it from your branch.  :)

So @stelleg , I believe this request can be closed, but I want to let you weigh in first.

stelleg:
@streichler sounds good to me

PR #286, opened 2017-08-14
Initial changes for physical tracing support

magnatelee:
This pull request contains the initial changes to support physical dependence analysis tracing. The current status is that mappers can choose in `select_task_options` for which tasks the runtime should memoize the mapping results (i.e. the output of `map_task`) and avoid calling the mappers again. This pull request is to be reviewed by my mentor at NVIDIA, Mike Bauer, who will check if this releases any sensitive IP.


magnatelee:
Commit 077b403 contains the first implementation of trace capturing procedure. The code collects a list of the Realm runtime calls and prints them on the screen.

magnatelee:
The feature to replay a single template has been implemented as of commit ee7e1c7.

runtime/legion.h
```diff
@@ -2524,6 +2524,7 @@ namespace Legion {
       Task(void);
     public:
       virtual const char* get_task_name(void) const = 0;
+      virtual bool has_trace(void) const = 0;
```
lightsighter: Is there a reason that we added this to the 'Task' object? I feel like this is something that should probably be in the mapping interface arguments and not something directly exposed in the 'Task' interface.

runtime/legion/garbage_collection.h
```diff
@@ -77,7 +77,8 @@ namespace Legion {
       PHYSICAL_MANAGER_REF = 20,
       LOGICAL_VIEW_REF = 21,
       REGION_TREE_REF = 22,
-      LAST_SOURCE_REF = 23,
+      PHYSICAL_TRACE_REF = 23,
```
lightsighter: Are we actually planning on making traces distributed collectable objects? If so, what is the use case? If not, there is a simpler garbage collection interface that we can use.

runtime/legion/legion_ops.cc
```diff
@@ -670,6 +693,7 @@ namespace Legion {
       // Tell our parent context that we are done mapping
       // It's important that this is done before we mark that we
       // are executed to avoid race conditions
+      //fprintf(stderr, "track_parent: %d, %p\n", track_parent, this);
```
lightsighter: Let's make sure we get rid of all these commented printfs before we do the final commit.

runtime/legion/legion_ops.cc
```diff
@@ -2365,6 +2411,8 @@ namespace Legion {
       if (remap_region)
       {
         region.impl->get_references(mapped_instances);
+        // TODO: Implement physical tracing for map operation
```
lightsighter: Let's aggregate the TODOs as github issue for a feature enhancement for physical tracing.

runtime/legion/legion_ops.cc
```diff
@@ -3158,6 +3213,7 @@ namespace Legion {
                                launcher.dst_requirements.size(), 
                              launcher.static_dependences,
                              launcher.predicate);
+      invoke_memoize_operation();
```
lightsighter: Are we sure we want to do these calls in the initialization phase of the operations and not later once they are in the pipeline?

runtime/legion/legion_ops.h
```diff
@@ -670,6 +688,22 @@ namespace Legion {
...
+     * physical analysis can be memoized.  Memoizable
+     * operations go through an extra step in the mapper
+     * to determine whether to memoize their physical analysis.
+     */
+    template<typename OP>
+    class MemoizableOp : public OP
+    {
+    public:
+      MemoizableOp(Runtime *rt);
+      virtual void execute_dependence_analysis(void);
```
lightsighter: Is this really all we need for a memoizable operation? I was expecting there to be more things. :)

lightsighter: For example, I feel like this could capture the 'invoke_memoize' method for most of the operation classes.

runtime/legion/legion_ops.inl
```diff
@@ -0,0 +1,38 @@
...
+
+    //--------------------------------------------------------------------------
+    template<typename OP>
+    void MemoizableOp<OP>::execute_dependence_analysis(void)
+    //--------------------------------------------------------------------------
+    {
+      OP::execute_dependence_analysis();
+    };
+  }; // namespace Internal
+}; // namespace Legion
```
lightsighter: I feel like we don't actually need this file since all the use cases of the template are in the source file where it is also defined so they will automatically be instantiated in the same translation unit. Let's just put these in legion_ops.cc for now (or legion_ops.h if you want). I would like to keep '.inl' files to ones that are needed for user facing APIs only.

runtime/legion.h
lightsighter: Is there a reason that we are making this part of the Task interface? It feels like something mapping specific and not something that should be visible to the application too. I feel like this should be exposed in the mapper interface and not here.

runtime/legion/legion_mapping.h
```diff
@@ -1440,6 +1441,29 @@ namespace Legion {
...
+       *
+       */
+      struct MemoizeInput {
+        bool traced;
+      };
+      struct MemoizeOutput {
+        bool memoize;
+      };
+      //------------------------------------------------------------------------
+      virtual void memoize_operation(const MapperContext  ctx,
```
lightsighter: This mapper call definitely needs some work. We need to have at least some context about the Mappable operation being traced.

runtime/legion/legion_ops.cc
```diff
@@ -291,6 +300,20 @@ namespace Legion {
     void Operation::execute_dependence_analysis(void)
     //--------------------------------------------------------------------------
     {
+#ifdef DEBUG_LEGION
+      assert(!memoizing || trace != NULL);
+#endif
+      if (memoizing)
```
lightsighter: Does this still belong here or should this be handled by the Memoizable trait now?

runtime/legion/legion_tasks.cc
```diff
@@ -3148,6 +3174,51 @@ namespace Legion {
...
+            if (acquired->find(manager) == acquired->end())
+              unacquired.push_back(manager);
+          }
+        }
+        if (Runtime::legion_spy_enabled)
+          runtime->forest->log_mapping_decision(unique_op_id, idx,
+                                                regions[idx],
+                                                instances);
+      }
+      runtime->forest->perform_missing_acquires(this, *acquired, unacquired);
```
lightsighter: Why is this necessary? You should really need this because I'm assuming your trace is already holding references to the physical instances to prevent them from being garbage collected and we shouldn't even be running through the code that checks the mapper output when we are replaying a trace.

runtime/legion/legion_tasks.cc
```diff
@@ -4203,6 +4305,22 @@ namespace Legion {
         deactivate();
     }
 
+    //--------------------------------------------------------------------------
+    void MultiTask::slice_index_space_for_replay(LegionTrace *trace)
```
lightsighter: This is alright for now, but we need to document some way that for distributed tracing we're going to need to do something different as making one slice that handles all the points won't really be practical.

runtime/legion/legion_tasks.cc
```diff
@@ -5760,6 +5912,13 @@ namespace Legion {
     void PointTask::perform_versioning_analysis(std::set<RtEvent> &ready_events)
     //--------------------------------------------------------------------------
     {
+      if (memoizing)
```
lightsighter: I feel like all these places where you have 'if (memoizing)' need to be captured by the Memoizable interface. We should talk about how to do this.

runtime/legion/legion_tasks.cc
```diff
@@ -8215,6 +8426,8 @@ namespace Legion {
       result->denominator = this->denominator * scale_denominator;
       result->index_owner = this->index_owner;
       result->remote_owner_uid = this->remote_owner_uid;
+      result->memoizing = memoizing;
+      result->trace_local_id = trace_local_id;
```
lightsighter: Things like this should be a member variable of the Memoizable class.

runtime/legion/legion_views.cc
```diff
@@ -545,7 +549,9 @@ namespace Legion {
                                                    const unsigned index,
                                                    const AddressSpaceID source,
                            LegionMap<ApEvent,FieldMask>::aligned &preconditions,
-                             std::set<RtEvent> &applied_events, bool can_filter)
+                                              std::set<RtEvent> &applied_events,
+                                                   bool tracing,
```
lightsighter: Why are we just passing a bool in all these cases and not the 'PhysicalTraceInfo' pointer (if there is one)?

runtime/legion/legion_views.cc
```diff
@@ -1550,7 +1618,65 @@ namespace Legion {
       }
       if (parent != NULL)
         parent->collect_users(term_events);
-    } 
+    }
+
+    //--------------------------------------------------------------------------
+    void MaterializedView::collect_traced_users()
```
lightsighter: I'm not really happy with this kind of architecture. I think views should be pretty ignorant of tracing and shouldn't really be responsible for remembering things about tracing. That's not really their job and I want them to be as unencumbered as possible since their multi-node implementations are already complex as it is.

runtime/legion/legion_views.cc
```diff
@@ -4327,7 +4509,8 @@ namespace Legion {
                               VersionTracker *src_version_tracker,
                   const LegionMap<ApEvent,FieldMask>::aligned &preconditions,
                         LegionMap<ApEvent,FieldMask>::aligned &postconditions,
-                        PredEvent guard, CopyAcrossHelper *across_helper) const
+                        PredEvent guard, PhysicalTraceInfo &trace_info,
```
lightsighter: Yes, this is what I was expecting to see everywhere that we have a 'bool tracing' in other method calls.

runtime/legion/legion_views.cc
```diff
@@ -7307,11 +7761,14 @@ namespace Legion {
...
           filter_local_users(*it);
       }
+      if (!traced_outstanding_gc_events.empty())
+        collect_traced_users();
 #if !defined(LEGION_SPY) && !defined(EVENT_GRAPH_TRACE) && \
       defined(DEBUG_LEGION)
       assert(reduction_users.empty());
       assert(reading_users.empty());
       assert(outstanding_gc_events.empty());
+      //assert(traced_outstanding_gc_events.empty());
```
lightsighter: Make sure we remove all commented code that we don't want.

runtime/legion/legion_views.cc
```diff
@@ -8151,6 +8697,60 @@ namespace Legion {
 #endif
     }
 
+    //--------------------------------------------------------------------------
+    void ReductionView::collect_traced_users()
```
lightsighter: Same thing here as for MaterializedView

runtime/legion/region_tree.cc
```diff
@@ -2481,15 +2514,33 @@ namespace Legion {
...
+      if (trace_info.tracing)
+      {
+#ifdef DEBUG_LEGION
+        assert(trace_info.tpl != NULL && trace_info.tpl->is_tracing());
+#endif
+        trace_info.tpl->record_merge_events(trace_info, result, result_events);
+      }
       // Return the merge of all the result events
-      return Runtime::merge_events(result_events);
+      return result;
```
lightsighter: This doesn't look right to me.

runtime/legion/region_tree.cc
```diff
@@ -2611,14 +2671,30 @@ namespace Legion {
...
-      return Runtime::merge_events(result_events);
+      ApEvent result = Runtime::merge_events(result_events);
+      if (trace_info.tracing)
+      {
+#ifdef DEBUG_LEGION
+        assert(trace_info.tpl != NULL && trace_info.tpl->is_tracing());
+#endif
+        trace_info.tpl->record_merge_events(trace_info, result, result_events);
+      }
+      return result;
```
lightsighter: Also doesn't look right

runtime/legion/legion_ops.cc
lightsighter: Ping on this again to make sure it gets done.

runtime/legion/legion_ops.cc
```diff
@@ -3813,21 +3883,44 @@ namespace Legion {
...
         }
         // Apply our changes to the version states
         // Don't apply changes to the source if we have a composite instance
         // because it is unsound to mutate the region tree that way
         if (src_composite == -1)
           src_versions[idx].apply_mapping(map_applied_conditions);
         dst_versions[idx].apply_mapping(map_applied_conditions); 
+#ifdef DEBUG_LEGION
+      dump_physical_state(&src_requirements[idx], idx);
+      dump_physical_state(&dst_requirements[idx], idx + src_requirements.size());
```
lightsighter: Do we really need to be dumping physical state in debug mode?

runtime/legion/legion_ops.cc
```diff
@@ -5444,7 +5606,7 @@ namespace Legion {
         case MAPPING_FENCE:
           {
             complete_mapping();
-            complete_execution();
+            complete_execution(Runtime::protect_event(execution_fence_event));
```
lightsighter: Why is this necessary? Mapping fences should only need to depend on the mapping of prior operations. It should have no dependence on the execution in any way.

runtime/legion/legion_ops.cc
```diff
@@ -6852,6 +7014,9 @@ namespace Legion {
 #ifdef DEBUG_LEGION
       assert(completion_event.exists());
       assert(chosen_instances.empty());
+#endif
+#ifdef DEBUG_LEGION
+      dump_physical_state(&requirement, 0, true, true);
```
lightsighter: More unnecessary dumping I think?

runtime/legion/legion_ops.cc
```diff
@@ -6944,6 +7109,9 @@ namespace Legion {
                                               chosen_instances);
       }
       version_info.apply_mapping(map_applied_conditions);
+#ifdef DEBUG_LEGION
+      dump_physical_state(&requirement, 0, false, true);
```
lightsighter: More unnecessary dumping I think?

runtime/legion/legion_ops.cc
```diff
@@ -8027,6 +8195,8 @@ namespace Legion {
       // Map this is a restricted region. We already know the 
       // physical region that we want to map.
       InstanceSet mapped_instances = restrict_info.get_instances();
+      // TODO: Implement physical tracing for acquire operation
```
lightsighter: Another acquire for the github issue list

runtime/legion/legion_ops.cc
```diff
@@ -8037,13 +8207,17 @@ namespace Legion {
...
+                                              NULL/*advance projections*/,
+                                              trace_info
 #ifdef DEBUG_LEGION
                                               , get_logging_name()
                                               , unique_op_id
 #endif
                                               );
       version_info.apply_mapping(map_applied_conditions);
+#ifdef DEBUG_LEGION
+      dump_physical_state(&requirement, 0);
```
lightsighter: More unnecessary dumping

runtime/legion/legion_ops.cc
```diff
@@ -8655,13 +8831,17 @@ namespace Legion {
...
```
lightsighter: unnecessary dumping...

runtime/legion/legion_ops.cc
```diff
@@ -12803,6 +13014,7 @@ namespace Legion {
         assert(value != NULL);
 #endif
         InstanceSet mapped_instances;
+        // TODO: Implement physical tracing for fill operation
```
lightsighter: TODO for the github list

runtime/legion/legion_ops.cc
```diff
@@ -12838,6 +13052,9 @@ namespace Legion {
                                         completion_event);
 #endif
         version_info.apply_mapping(map_applied_conditions);
+#ifdef DEBUG_LEGION
+      dump_physical_state(&requirement, 0);
```
lightsighter: last one of these I'll annotate, you can find the rest.

runtime/legion/legion_ops.cc
```diff
@@ -12889,6 +13106,8 @@ namespace Legion {
       void *result = malloc(result_size);
       memcpy(result, future.impl->get_untyped_result(), result_size);
       InstanceSet mapped_instances;
+      // TODO: Implement physical tracing for fill operation
```
lightsighter: Last TODO I will annotate as well

runtime/legion/legion_tasks.cc
lightsighter: Pinging you on this again since I don't think you've addressed it yet

runtime/legion/legion_views.h
```diff
@@ -804,25 +836,29 @@ namespace Legion {
...
-                                 std::set<RtEvent> &applied_events);
+                                 std::set<RtEvent> &applied_events,
+                                 PhysicalTraceInfo &trace_info);
       virtual ApEvent find_user_precondition(const RegionUsage &user,
                                            ApEvent term_event,
                                            const FieldMask &user_mask,
                                            Operation *op, const unsigned index,
                                            VersionTracker *version_tracker,
-                                           std::set<RtEvent> &applied_events);
+                                           std::set<RtEvent> &applied_events,
```
lightsighter: Why is the applied_events now needed on some of these methods?

runtime/legion/runtime.cc
```diff
@@ -18003,6 +18049,9 @@ namespace Legion {
           if (!strcmp(argv[i],"-hl:safe_mapper"))
             unsafe_mapper = false;
           BOOL_ARG("-hl:inorder",program_order_execution);
+          BOOL_ARG("-hl:dump_physical_traces",dump_physical_traces);
+          BOOL_ARG("-hl:no_tracing",no_tracing);
+          BOOL_ARG("-hl:no_physical_tracing",no_physical_tracing);
```
lightsighter: I don't think you need to add -hl versions of these flags. The old ones are just there for backwards compatibility.

runtime/legion.h
```diff
@@ -2444,6 +2444,36 @@ namespace Legion {
...
+        FILL_MEMOIZABLE,
+        DYNAMIC_COLLECTIVE_MEMOIZABLE,
+      };
+      virtual MemoizableType get_memoizable_type(void) const = 0;
+      virtual const Task* as_task(void) const = 0;
+      virtual const Copy* as_copy(void) const = 0;
+      virtual const Fill* as_fill(void) const = 0;
+      virtual const DynamicCollective* as_dynamic_collective(void) const = 0;
+    };
+
```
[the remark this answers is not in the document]
magnatelee: I would have used the existing Mappable interface if DynamicCollectOp were a mappable operation. I'm fine with switching this to Mappable.

runtime/legion.h
```diff
@@ -5453,7 +5483,7 @@ namespace Legion {
        * The trace ID need only be local to the enclosing context.
        * Traces are currently not permitted to be nested.
        */
-      void begin_trace(Context ctx, TraceID tid);
+      void begin_trace(Context ctx, TraceID tid, bool memoize = true);
```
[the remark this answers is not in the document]
magnatelee: I agree. Let me change it to sound more meaningful.

runtime/legion/legion_ops.cc
magnatelee: Well, this isn't my code, but the code that existed in the snapshot I took. The dump_physical_state function returns immediately if the user did not request the tree dump. I guess an alternative is to hoist that condition to every place we call this function. If you prefer that way, I can make changes.

runtime/legion/legion_ops.cc
magnatelee: If that's the intended semantics, I will revert this change.

runtime/legion/legion_ops.cc
magnatelee: I think some of these comments do not make sense any more. I'll change them to a list of operations the tracing does not (and will not) support.

runtime/legion/legion_ops.cc
magnatelee: I don't think this comment is even true anymore.

runtime/legion/legion_views.h
magnatelee: What do you mean? I didn't add applied_events to this function.

runtime/legion/runtime.cc
magnatelee: I'll remove them.

PR #287, opened 2017-08-16
Added ability to specify a custom HDF5 installation path

akshaysubr:
HDF_ROOT environment variable that points to the HDF5 installation directory is used if present.


elliottslaughter:
Seems reasonable. Let's wait for the tests to pass, then we'll pull it.

elliottslaughter:
Failure is unrelated. Going to pull.

PR #297, opened 2017-09-15
Added OpenMP logic to cmake

ipdemes:


streichler:
@ipdemes Can you also add an option for -DREALM_OPENMP_KMP_SUPPORT?  Ideally, both GOMP and KMP support would default to ON if REALM_USE_OPENMP is enabled.  (This is how the Makefile version works.)

ipdemes:
@streichler : I have just added the logic for the REALM_OPENMP_KMP_SUPPORT

ipdemes:
@streichler : Oops, I forgot to remove it, sorry. It should be good now.

PR #298, opened 2017-09-18
Attach external array to logical region

eddy16112:


runtime/realm/inst_layout.h
```diff
@@ -144,7 +144,7 @@ namespace Realm {
     bool serialize(S& serializer) const;
 
     ZPoint<N, size_t> strides;
-    size_t offset;
+    ptrdiff_t offset;
```
streichler: @eddy16112 I know it looks weird, but this does need to be `size_t`.  Over/underflow is not defined for signed integers (dumb, but true).

streichler:
The realm side of this looks fine, except for the one issue I flagged.  My thinking is that the example programs ought to go in examples rather than tutorials, but I could be talked into having one array-attachment tutorial as well.

runtime/realm/inst_layout.h
eddy16112: what if the offset is negative? I know the size_t is able to handle this kind of underflow, and somehow the entire program still works, but I am not sure if it will bring any invisible bugs.

streichler: A negative offset becomes a very large size_t value which, when added to another size_t value, wraps around 2^64 correctly and results in a sum that is the right amount smaller.

streichler:
@eddy16112 The CI tests on Travis are flagging some issues that need to be cleaned up.  Here's one:
```/home/travis/build/StanfordLegion/legion/runtime/legion/legion_ops.cc: In member function ‘virtual void Legion::Internal::DetachOp::trigger_mapping()’:

/home/travis/build/StanfordLegion/legion/runtime/legion/legion_ops.cc:14365:24: error: unused variable ‘inst_manager’ [-Werror=unused-variable]

       InstanceManager *inst_manager = manager->as_instance_manager(); ```

A second issue looks to be a problem with the cmake build - if you added a source file, it needs to be added to runtime/CMakeLists.txt as well I believe.

**Comment by elliottslaughter** [comment]
@streichler @eddy16112 Wherever the tests are placed I'd like them to be added to the appropriate section of the test suite (see below). All code in the Legion repo must be actively tested.

https://github.com/StanfordLegion/legion/blob/master/test.py#L39-L62

File: runtime/legion/region_tree.cc [hunk 2]
```diff
@@ -7943,7 +7943,7 @@ namespace Legion {
       }
 #ifdef DEBUG_LEGION
       assert(layout != NULL);
-      assert(layout->constraints->specialized_constraint.is_file());
+   //   assert(layout->constraints->specialized_constraint.is_file());
```
eddy16112: I don't know how to handle this assert properly. For external array, I just set the SpecializedConstraint(NORMAL_SPECIALIZE), so there is no way for me check it.

eddy16112:
@elliottslaughter Done with the test.py. Thanks for your comments

lightsighter:
All the examples do need to go in the 'examples' directory and not the 'tutorial' directory because 'tutorials' are exclusively reserved for things on the actual Legion webpage.

I'm concerned about some of the lines that were commented out checking for errors. We need to handle those properly.

The changes in the Legion runtime need to be made to match the Legion style:
- All braces get there own line
- There is a strict 80 character limit on each line
- ...

streichler:
I managed to accidentally close this by deleting the deppart branch.  Deppart changes are in master, so this should target master now.

streichler:
@eddy16112 It looks like github's textual merge of legion_ops.cc isn't good enough: https://travis-ci.org/StanfordLegion/legion/jobs/284337824
Please clean this up, and also resolve the conflict on language/src/rdir.  This is a submodule, so you need to make sure that subtree is sync'd to the same version as master (https://github.com/StanfordLegion/rdir/commit/2c0b613b6ebe990de2c890ec81d819daae5ac440) and then add/commit it to your branch.

PR #299, opened 2017-09-22
Fixes to setup_env.py

manopapad:


elliottslaughter:
I made some changes to give better error messages when setup_env.py fails. If there is anything else you'd like to see, please let me know.

PR #303, opened 2017-09-29
Regent pretty-printer fixes

manopapad:


elliottslaughter:
There's one minor issue with this change set, which I am fixing, and then I will merge.

PR #306, opened 2017-10-02
Python support

elliottslaughter:
This introduces basic support for Python via a new Realm module and native Python APIs for Legion.


streichler:
Two minor things:
1) I think we can probably leave out gen_dlmproxy.py - we're pretty sure that path isn't going to work.
2) Do you want to take a shot at supporting python in the cmake build?

Issue #257, opened 2017-05-22
Some header names are too generic


junghans:
With an legion installed in `/usr` or `/usr/local`, one can make legion pick up the wrong header:
```
$ cat accessor.h
#error Wrong header
$ cat main.cpp
#define LEGION_CMAKE
#include <legion.h>

int main () {
  return 1;
}
$ g++ -I. -I/usr/include/legion -I/usr/include/realm main.cpp
In file included from /usr/include/realm/redop.h:22:0,
                 from /usr/include/realm/event.h:23,
                 from /usr/include/realm/processor.h:23,
                 from /usr/include/realm/profiling.h:27,
                 from /usr/include/realm/realm.h:23,
                 from /usr/include/legion/legion_types.h:43,
                 from /usr/include/legion/legion.h:36,
                 from main.cpp:2:
./accessor.h:1:2: error: #error Wrong header
 #error Wrong header
  ^~~~~
```

Looking at what legion installs:
```
/usr/include/accessor.h
/usr/include/activemsg.h
/usr/include/arrays.h
/usr/include/atomics.h
/usr/include/channel.h
/usr/include/channel_disk.h
/usr/include/utilities.h
```
`accessor.h`, `arrays.h`, `atomics.h` and `utilities.h` might be too generic names and should be moved to a subdirectory.

lightsighter:
Any header files you see directly in the runtime/ directory are going to be dead in the next few months (including all the ones you see above).  Users should never need to include anything other than realm.h or legion.h by the time we are done.

streichler:
The header_cleanup branch (initial commit fbe7ab6) moves all the real headers out of the base runtime/ directory.  For compatibility with older apps using the older Makefiles, there are still files there that redirect (with a warning) to the new locations, but cmake should not be installing them.

Rather than changing the names of header files that might conflict with other names, my plan is to make all #include paths within Legion and Realm be rellative to the root (i.e. explicitly include legion/ or realm/ in the path).  That'll come next.

streichler:
All Legion and Realm headers are now #include'd using their full path (from the runtime root directory), which should eliminate this issue.  @junghans , when you get a chance, can you take a look and see if you agree the changes are sufficient?

streichler:
@junghans, do you have any remaining concerns?  If not, I'll close this issue.

PR #323, opened 2017-10-22
move all headers out of runtime base directory

streichler:
Fixes for #257.  Adding it as a pull request to get Travis to test it.

### DIFF
--- a/runtime/legion.h
+++ b/runtime/legion.h
@@ -13,12 +13,10 @@
  * limitations under the License.
  */
 
-
-#ifndef __LEGION_RUNTIME_H__
-#define __LEGION_RUNTIME_H__
-
-// decide whether we want C and/or C++ bindings (default matches
-//  host language)
+// decide whether we want C and/or C++ bindings (default matches host language)
+//
+// each set of bindings has its own include-once ifdef armor, allowing the
+//  second set of bindings to be loaded even if the first already has been
 #if !defined(LEGION_ENABLE_C_BINDINGS) && !defined(LEGION_DISABLE_C_BINDINGS)
   #ifndef __cplusplus
     #define LEGION_ENABLE_C_BINDINGS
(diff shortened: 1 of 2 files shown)


streichler:
@elliottslaughter no need to look at all the changes to legion and realm include files, but if you could look over the handful of changes in language/ and bindings/, I'd appreciate it.

bindings/terra/bishop_c.h
```diff
@@ -16,7 +16,8 @@
 #ifndef __BISHOP_C_H__
 #define __BISHOP_C_H__
 
-#include "legion_c.h"
+#define LEGION_ENABLE_C_BINDINGS
+#include "legion.h"
```
elliottslaughter: My first concern with this approach is that if someone does:

```
#include "legion.h"
#include "bishop_c.h"
```

Then on the first include of `legion.h` the define won't be set, and then the include will be a nop on the second include. Is `leigon.h` set up to mitigate this?

My second concern is that since this is called `bishop_c.h` it is presumably a pure-C file. We should check with @magnatelee to be sure, but it really shouldn't be named `*_c.h` if it isn't. If so, then it would be really really bad to include any C++ functionality. It should really just include the C API directly.

bindings/terra/legion_terra.h
```diff
@@ -16,7 +16,7 @@
 #ifndef __LEGION_TERRA_H__
 #define __LEGION_TERRA_H__
 
-#include "legion_c.h"
+#include "legion.h"
```
elliottslaughter: Again, same as before. This is externally facing pure-C functionality. Do not include C++ headers from this file.

bindings/terra/legion_terra_partitions.h
```diff
@@ -24,12 +24,13 @@
 //
 // ******************** IMPORTANT **************************
 
+#define LEGION_ENABLE_C_BINDINGS
+#include "legion.h"
```
elliottslaughter: Same. In fact, I even wrote a comment about this above.

I'm going to stop commenting on this here, but this issue seems to recur throughout.

PR #339, opened 2017-12-18
Deterministic compilation for Regent

### LINKED ISSUE #334: Regent: non-determinism in code generation
Regent currently produces code non-deterministically in a way that causes dynamic control replication to fail. Ideally, at least the sequence of generated runtime calls would be deterministic.

elliottslaughter:
Making this a pull request to track test status.

Fixes #334.

@magnatelee We still need you to update the parallelizer.


magnatelee:
Commit b4552b9 fixes the crash in the parallelizer.

PR #357, opened 2018-02-20
Compile tasks in parallel

manopapad:
Requires https://github.com/zdevito/terra/pull/269


language/src/regent/std.t
```diff
@@ -3435,11 +3433,72 @@ function std.setup(main_task, extra_setup_thunk, registration_name)
...
+  for _,variant in ipairs(variants) do
+    variant.task:complete()
+  end
+
+  -- Spawn children processes, assign a subset of tasks to each to compile.
+  -- HACK: Terra functions used by more than one task may get compiled
+  -- multiple times, and included in multiple object files by different
+  -- children. They should all be exactly the same, so the linker may choose
+  -- to keep any one of them.
+  local flags = terralib.newlist({'-Wl,--allow-multiple-definition'})
```
elliottslaughter: Is this still needed with the Terra change?

Edit: Is fixed in a later commit. Would be nice to reorganize the patches to avoid this.

language/tests/regent/run_pass/paral_saveobj.rg
```diff
@@ -0,0 +1,40 @@
+-- Copyright 2018 Stanford University
```
elliottslaughter: Please name this file in complete words e.g. `compile_parallel_saveobj.rg`. Same for the other test.

language/src/regent/std.t
```diff
@@ -3150,38 +3150,6 @@ function std.register_variant(variant)
   variants:insert(variant)
 end
 
-local function make_task_wrapper(task_body)
```
elliottslaughter: I'm fine with changing the public interface to `variant:make_wrapper()` as in a later commit, but personally I'd like to keep this function as it is (just move it to the other file, immediately before the definition of `variant:make_wrapper()`) because it's such a clear, well-defined interface. This is code I can copy-and-paste anywhere and have be useful, while the method version needs to be disentangled from its context.

elliottslaughter:
Also, there's something going wrong with the Legion Spy tests that we should resolve before merging. (See the linked Travis failures.)

language/src/regent/std.t
manopapad: Done

language/tests/regent/run_pass/paral_saveobj.rg
manopapad: Done

manopapad:
I believe the Spy test is failing because that testcase doesn't call start(), instead it does saveobj and runs the executable using os.execute. There doesn't seem to be other tests that do this, so I don't know how to fix this for Spy.

As for reducing the amount of changes, are you suggesting I reset to the beginning of the branch and reapply the changes?

manopapad:
Updated to do work-stealing, to handle load imbalance in compilation.

elliottslaughter:
Re: the Legion Spy issue, you're missing the arguments when you execute the saved object. This is how we do it currently in the test suite:

https://github.com/StanfordLegion/legion/blob/stable/language/examples/embed.rg#L62-L69

elliottslaughter:
Looks like the Mac builds are unhappy. You can't rely on importing `stderr`, you need to open the file descriptor like this:

https://github.com/StanfordLegion/legion/blob/master/language/src/regent/std_base.t#L44

language/travis.py
```diff
@@ -47,6 +47,7 @@ def test(root_dir, install_only, debug, short, spy, gcov, hdf5, openmp, env):
         if gcov: extra_flags.append('--run')
         if hdf5: extra_flags.append('--hdf5')
         if openmp: extra_flags.append('--openmp')
+        extra_flags.extend(['--extra=-fjobs', '--extra=' + jobs])
```
elliottslaughter: Probably want `'--extra=%s' % jobs`

manopapad:
The tests involving Bishop will still fail, until https://github.com/zdevito/terra/pull/271 is merged. This is happening because Terra/LLVM handles some of the constants that Bishop declares (using terralib.constant) by emitting external-linkage entries on the BSS segment, all named "<constant>", thus causing a name conflict during linking if two different processes both need to emit a constant.

PR #363, opened 2018-03-08
Missed typo from #357

manopapad:


elliottslaughter:
I think this was already fixed by https://gitlab.com/StanfordLegion/legion/commit/a34095285e867027487798729f4fc2968560ec7b. Does that commit look ok?

manopapad:
Yeah that's actually a better fix.

PR #364, opened 2018-03-09
Support compilation with CUDA on head nodes w/o a GPU

manopapad:
When building with CUDA on a head node w/o a GPU, libcuda.so will not be found
in /usr/lib, but a stub will exist in $CUDA_HOME/lib64/stubs. We need to add an
explicit pointer to that file when building Regent (in runtime.mk), and also
when linking a Regent program with libregent.so (in regent.py).

The final executable will not contain a direct pointer to this stub libcuda.so.
It will instead list a dependency on SONAME=libcuda.so.1. When we actually run
it on a node with a GPU, the real libcuda.so library will exist in /usr/lib, and
the dynamic linker should be able to locate it.

### DIFF
--- a/language/regent.py
+++ b/language/regent.py
@@ -79,6 +79,7 @@ if os_name == 'Darwin':
 lib_path = (
     (os.environ[LD_LIBRARY_PATH].split(':')
      if LD_LIBRARY_PATH in os.environ else []) +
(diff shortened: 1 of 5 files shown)


manopapad:
Actually ignore this PR, apparently even nodes w/o a GPU that have the CUDA SDK installed should have libcuda.so in /usr/lib64. It just happened that one of the login nodes on the cluster I was using (Certainty) was misconfigured.

Issue #369, opened 2018-03-20
Memory corruption when running with non-thread-safe libhdf


manopapad:
Currently, for applications that use HDF, the user code needs to enter libhdf to set up datasets, create files etc. The DMA processor also needs to enter libhdf, to actually perform the I/O. This means it's possible for a CPU processor and a DMA processor to be in libhdf at the same time. By default, libhdf is not thread-safe, so this situation can result in memory corruption. One way to work around this is to simply always link Legion with a libhdf that was compiled to be thread-safe (`./configure --enable-threadsafe --disable-hl`).

lightsighter:
It looks like someone started to add code to Realm's HDF5 module to check to see if it supported thread safety or not and then do the right thing if it was not thread-safe, but it doesn't look like the results of this check are used anywhere:
https://github.com/StanfordLegion/legion/blob/master/runtime/realm/hdf5/hdf5_module.cc#L90-L102
@streichler @jiazhihao Can you say what Realm should do with a non thread-safe implementation of HDF5?

streichler:
That detection code was added so that we could decide whether it was ok for the dma system to do HDF5 transfers on multiple internal dma threads at a time.  The new issue here is interactions between the dma system and application code, and it's not clear that there'll be a pretty answer to it.

PR #370, opened 2018-03-21
Add HDF installation to setup_env.py

manopapad:
Built in thread-safe mode, so that Regent users at least won't encounter #369.

### DIFF
--- a/language/scripts/setup_env.py
+++ b/language/scripts/setup_env.py
@@ -162,7 +162,6 @@ def build_regent(root_dir, use_cmake, cmake_exe,
         ('GASNET', gasnet_dir),
         ('USE_GASNET', os.environ['USE_GASNET'] if 'USE_GASNET' in os.environ else '1'),
         ('LLVM_CONFIG', os.path.join(llvm_dir, 'bin', 'llvm-config')),
-        ('USE_HDF', os.environ['USE_HDF'] if 'USE_HDF' in os.environ else '1'),
         ('HDF_ROOT', hdf_dir),
     ])
 
--- a/language/scripts/setup_env.py
+++ b/language/scripts/setup_env.py
@@ -138,13 +138,32 @@ def build_terra(terra_dir, llvm_dir, cache, is_cray, thread_count):
         cwd=terra_dir,
         env=env)
 
+def build_hdf(source_dir, install_dir, thread_count, is_cray):
+    env = None
+    if is_cray:
+        env = dict(list(os.environ.items()) + [
+            ('CC', os.environ['HOST_CC']),
+            ('CXX', os.environ['HOST_CXX']),
+        ])
+    subprocess.check_call(
+        ['./configure',
+         '--prefix=%s' % install_dir,
+         '--enable-threadsafe',
+         '--disable-hl'],
+        cwd=source_dir,
+        env=env)
+    subprocess.check_call(['make', '-j', str(thread_count)], cwd=source_dir)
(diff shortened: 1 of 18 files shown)


language/scripts/setup_env.py
```diff
@@ -138,13 +138,32 @@ def build_terra(terra_dir, llvm_dir, cache, is_cray, thread_count):
         cwd=terra_dir,
         env=env)
 
+def build_hdf(source_dir, install_dir, thread_count, is_cray):
+    env = None
+    if is_cray:
```
elliottslaughter: Do we need this part? I think this was something we were doing for Terra because otherwise Terra wouldn't build properly, but I would expect HDF to build alright with the default system compiler.

language/scripts/setup_env.py
```diff
@@ -138,13 +138,32 @@ def build_terra(terra_dir, llvm_dir, cache, is_cray, thread_count):
...
+
 def build_regent(root_dir, use_cmake, cmake_exe,
-                 gasnet_dir, llvm_dir, terra_dir, conduit, thread_count):
+                 gasnet_dir, llvm_dir, terra_dir, hdf_dir, conduit, thread_count):
     env = dict(list(os.environ.items()) + [
         ('CONDUIT', conduit),
         ('GASNET', gasnet_dir),
         ('USE_GASNET', os.environ['USE_GASNET'] if 'USE_GASNET' in os.environ else '1'),
         ('LLVM_CONFIG', os.path.join(llvm_dir, 'bin', 'llvm-config')),
+        ('USE_HDF', os.environ['USE_HDF'] if 'USE_HDF' in os.environ else '1'),
```
elliottslaughter: I'm not sure we want this enabled by default. I'm guessing we could probably let the user turn this on with a flag like `--hdf` or something. In which case this conditional would turn into `os.environ['USE_HDF'] if 'USE_HDF' in os.environ else ('1' if use_hdf else '0')`

manopapad: Actually I'd rather remove line 165 altogether, and avoid adding another option to setup_env.py. This will cause install.py to read the value of USE_HDF from the environment directly. In that scipt, HDF is off by default, unless USE_HDF=1.

manopapad:
One thing I forgot to note is, now when running the Regent compiler (and/or a saveobj-ed binary), the user needs to add language/hdf/install/lib to LD_LIBRARY_PATH manually. We could avoid this by including libhdf.a into libregent.so if we wanted.

language/scripts/setup_env.py
manopapad: I'll try it on Titan and see if it works without this block.

language/scripts/setup_env.py
elliottslaughter: That's fine.

language/scripts/setup_env.py
manopapad: This block is necessary when building on Titan. During compilation of the HDF library, the build process needs to run an executable (H5Detect) that's produced at an earlier point, and without this block H5Detect contains instructions that are illegal for the head node.

language/scripts/setup_env.py
manopapad: Done

language/scripts/setup_env.py
elliottslaughter: Fun.

elliottslaughter:
Thanks, pushed to gitlab. Will show up as merged here at some point....

PR #391, opened 2018-05-15
Doctor: region creation without deletion, overlapping inputs

debnil:
- create a region without delete
- check for parameters with overlapping regions
- tests added for extra privileges + overlapping regions


elliottslaughter:
Please update the pull request to target the master branch. Also, if there are changes to RDIR please make sure there is a separate pull request for that repo. (It's possible that RDIR appeared to change only because you are issuing this PR to stable.)

Thanks!

PR #399, opened 2018-06-03
legion.cc: fix potential return of random data

junghans:
/cc @nmorey

### DIFF
--- a/runtime/legion/legion.cc
+++ b/runtime/legion/legion.cc
@@ -4075,6 +4075,7 @@ namespace Legion {
                     "replaced with a call to create_partition_by_field.",
                     ctx->get_task_name(), ctx->get_unique_id());
       assert(false);
+      return IndexPartition::NO_PART;
     }
 
     //--------------------------------------------------------------------------


streichler:
I'll let @lightsighter look over the change and do the merge, but as a note, we do not recommend building Legion or Realm with -DNDEBUG.  We use `assert(0)` instead of `exit(1)` for fatal errors so that trapping them a debugger is easier, and defining `NDEBUG` causes those to fall through/return to application code instead.

elliottslaughter:
Could we `abort()` instead of `assert(false)`? That should have the same behavior w.r.t. trapping in a debugger, and won't subject us to the whims of whether `NDEBUG` is set or not.

junghans:
`abort()` would be better!

lightsighter:
I'm fine with the patch. I'm not making any more pervasive changes like changing the `assert(false)` to `abort` until @magnatelee finishes with the merge of the 'memoize' branch due to the risk of creating massive number of conflicts.

streichler:
I've created issue #400 to make sure we remember to do the assert->abort switch.

PR #443, opened 2018-10-18
Fix value returned from RectInDomainIterator::operator->()

mpokorny:
Return value expression was missing address-of operator


streichler:
Thanks @mpokorny!  I've cherry-picked this into our master branch as commit c6eda4b - it
should show up on github as soon as the CI gods bless it.

PR #462, opened 2018-12-19
Fix "IndexPartition" misspelling

mpokorny:


mpokorny:
Would you prefer that I submit simple PRs like this one to the master branch in the future?

streichler:
Master's slightly better than stable, but only if there's merge conflicts for the submitter to clean up.  Any pull request submitted actually gets applied to a repo on gitlab that's then mirrored over here for CI goodness reasons.

PR #544, opened 2019-06-07
regent: add optional disjointness for image

lizdulac:


language/src/regent/type_check.t
```diff
@@ -4198,8 +4200,7 @@ function type_check.top_fspace(cx, node)
   }
 end
 
-function type_check.top_quote_expr(cx, node)
-  -- Type check lazily, when the expression is interpolated.
+function type_check.top_quote_expr(cx, node)  -- Type check lazily, when the expression is interpolated.
```
elliottslaughter: Looks like this newline got deleted by accident?

language/tests/regent/compile_fail/region_partition_by_aliased_image.rg
```diff
@@ -0,0 +1,80 @@
...
+      @x *= i + 2
+    end
+  end
+
+  var t = 0
+  for x in r do
+    t += @x
+  end
+
+  assert_disjoint(q[0], q[1])
```
elliottslaughter: Could we name these with variables (e.g. `var q0 = q[0]` and `var q1 = q[1]`)? Otherwise this will break on any changes to the compiler's internal region numbering. (Similarly for other cases of unnamed subregions in the compile-fail tests.)

elliottslaughter:
Looks like some of the tests that got added are not actually disjoint, so we'll need to fiddle with them to make sure that they are:

https://travis-ci.org/StanfordLegion/legion/jobs/543993196

Otherwise looks good modulo the one oustanding comment above.

elliottslaughter:
This has been merged (though Github doesn't know because I squashed the commits before pushing). Thanks!

Issue #575, opened 2019-08-25
hwloc-2.0 support


keszybz:
Hi,
mpich 3.3 now requires hwloc>=2.0. It would be great if legion could be updated too.
This was already signalled in #479 in February :(

streichler:
Yes, this is somewhere on our task list.  It's not super-high though, as you can almost
certainly disable hwloc in the Legion/Realm build with no loss of functionality.  (Or
conversely, if you have a system that the linux-sysfs-based discovery code doesn't work
as well as hwloc for, let me know and we'll consider bumping this up the priority list.)

keszybz:
Thanks for the quick reply. I made a small patch to fix one of the two compilation issues.
I also did the suggested thing and disabled the use of hwloc that does not compile...

> Or conversely, if you have a system that the linux-sysfs-based discovery code doesn't work
as well as hwloc for, let me know and we'll consider bumping this up the priority list.

I wish I knew ;) I'm just the packager here, and I'm only touching legion because I want to update mpich.

streichler:
As mentioned in the pull request, I'd like this change to add hwloc-2.x support without losing hwloc-1.x support, so it'll all need to be ifdef-protected.

scardracs:
Still waiting for update to hwloc-2.6.0

scardracs:
If you have any patch please provide it as I could add it to gentoo ebuild

streichler:
@scardracs As was suggested a while back, I think the simplest thing to do here is to remove the Legion package's dependence on hwloc entirely.  We routinely build without hwloc in our development, so I'm not sure what the issue that @keszybz encountered before was.  If you're not sure how to disable use of hwloc in Legion, please point me at your build script and I'll see if I can suggest something.

scardracs:
> @scardracs As was suggested a while back, I think the simplest thing to do here is to remove the Legion package's dependence on hwloc entirely.  We routinely build without hwloc in our development, so I'm not sure what the issue that @keszybz encountered before was.  If you're not sure how to disable use of hwloc in Legion, please point me at your build script and I'll see if I can suggest something.

Yeah I know how to remove it, the only point is that Gentoo ebuild use +hwloc, which means is forced by default. If you prefer to remove it I can proceed on dropping it on Gentoo side (as I'm working on cleanup/maintain it)

streichler:
I'm not that familiar with Gentoo's build system, but having Legion's use of hwloc be disabled by default sounds better to me.

[...]

PR #576, opened 2019-08-25
hwloc-2.0: use hwloc_linux_read_path_as_cpumask()

keszybz:
See https://github.com/open-mpi/hwloc/commit/e7f5962f014bbf9dcd284498b8662ec7645c2d4e.

The easy part of #575.


keszybz:
This probably shouldn't be merged until the other issue is fixed and hwloc dependency is bumped, as it will not compile with hwloc-1.8.

streichler:
Thanks for taking a shot at this!

Does hwloc 2.x have ifdefs that distinguish it from hwloc 1.x?  I don't want to lose hwloc 1.x support as part of this.

keszybz:
It has `#define HWLOC_API_VERSION 0x00020000` now.

PR #583, opened 2019-09-14
Allow assignment operations on regions in Regent

ldr709:
This is not ready for merge. Issues:

Doesn't support `__fields` or `__physical`. It would need to know the privileges on the fields during type checking, but that information is now calculated later. Breaking the circular dependency probably wouldn't be too difficult, but it would be annoying to have to go back and fix up the type checked AST after finding the privileges. One failing test case is `run_pass/call_terra_region.rg`.

Index launch checking is less complete. I didn't get around to implementing the check to see if two regions are always distinct for all pairs of iterations. This would just need some metadata to tell if a parent region was declared outside the loop. A failing test case is `run_pass/optimize_index_launch_num.rg`.

`flow_from_ast.t` still tries to use the old constraint data, even though it isn't there anymore. It hasn't been converted to using the new interface for getting constraint information. This makes `run_pass/parallelize_tasks_hint_color_space.rg` fail, among others.

### DIFF
--- a/language/src/common/data.t
+++ b/language/src/common/data.t
@@ -315,6 +315,7 @@ data.vector.__index = data.vector
 
 function data.vector.__eq(a, b)
   assert(data.is_vector(a) and data.is_vector(b))
+  -- FIXME: Looks like a bug, as it ought to make sure they are the same length.
   for i, v in ipairs(a) do
     if v ~= b[i] then
       return false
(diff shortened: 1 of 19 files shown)


streichler:
@elliottslaughter @magnatelee what's the plan for this?

elliottslaughter:
I think @magnatelee looked at this and I believe our concern was about compile times. I think it's still very desirable feature but just haven't had time to look in detail at what the impact on compile times is and whether it can be mitigated.

PR #585, opened 2019-09-14
Type checking for functions returning regions in Regent

ldr709:
All permissions and constraints on return values are automatically inferred. It doesn't work for mutually recursive tasks returning regions, though it does for normal recursions. It might be useful to either fix this or allow them to be specified in the task signature.

Code generation for handling the return values is missing, so it will trigger an assertion failure in `codegen.t` when it correctly type checks a region-returning function call.


streichler:
@elliottslaughter @magnatelee what's the plan for this?

elliottslaughter:
This depends on #583, and therefore has a similar status.

PR #658, opened 2019-10-11
Print expected number of fields.

syamajala:
Print the expected number of fields and how many fields a c/c++ launcher received.

### DIFF
--- a/language/src/regent/header_helper.t
+++ b/language/src/regent/header_helper.t
@@ -442,10 +442,11 @@ local function make_add_argument(launcher_name, wrapper_type, state_type,
     local field_paths, _ = base.types.flatten_struct_fields(param_type:fspace())
     arg_setup:insert(
       quote
-          var msg : int8[128]
-          c.sprintf(msg, [launcher_name .. " wrong number of fields for region " .. tostring(arg_value) .. " (argument " .. param_i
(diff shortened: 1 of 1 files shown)


language/src/regent/header_helper.t
```diff
@@ -442,8 +442,10 @@ local function make_add_argument(launcher_name, wrapper_type, state_type,
     local field_paths, _ = base.types.flatten_struct_fields(param_type:fspace())
     arg_setup:insert(
       quote
-          base.assert([field_count] == [#field_paths],
-            ["wrong number of fields for region " .. tostring(arg_value) .. " (argument " .. param_i .. ")"])
+          var msg : int8[128]
+          c.sprintf(msg, [launcher_name .. " wrong number of fields for region " .. tostring(arg_value) .. " (argument " .. param_i
+                            .. ") expected: " .. #field_paths .. " got: %d"], field_count)
+          base.assert([field_count] == [#field_paths], msg)
```
elliottslaughter: Instead of using `assert` for this, let's use an `if` to check the condition followed by `printf` and `abort` if it's false.

syamajala: This should now be fixed.

elliottslaughter:
Merged via e4b4847f1038bb2ddad3a461ebf0a81958477778 into master. Thanks!

PR #677, opened 2019-11-05
cmake: try using perl for portable source file generation

streichler:


streichler:
PR was to get Travis to test it.  Things look good, so closing this PR and will merge through the normal path.

PR #763, opened 2020-02-27
Add completeness keyword

ndrewtl:
Adds completeness keyword for partitioning. Similar to disjointness, allows partitions to specify that they cover the entire region, which can then be passed down to the runtime for optimizations.


elliottslaughter:
This is merged as of bb90442bcd76cf985c534e77f914a9858d5f1690. (Github doesn't see this, since I rebased in the process, but it's all in there now.)

Thanks @ndrewtl again for your work on this!

Issue #817, opened 2020-04-07
Regent: loop fails with ispace.bounds.hi


rupanshusoi:
The following

```
import "regent"

task foo(r : region(ispace(int1d), int))
  where reads writes(r) do
end

task main()
  var r = region(ispace(int1d, 10), int)
  fill(r, 1)
  var is = ispace(int1d, 5)
  var p = partition(equal, r, is)
  for i in is do
    foo(p[i + is.bounds.hi])
  end
end
regentlib.start(main)
```
fails with
```
regent/codegen.t:806: definition of variable with symbol $bounds_ispace(int1d) is not in scope in this context
    var [value_name] = [value]
                            ^
```
The error is fixed by introducing a new variable

```
  var c = is.bounds.hi
  for i in is do
    foo(p[i + c])
  end
```

I realise the partition access is going out of bounds in both cases, but the compiler usually doesn't complain about it. Fixing this also seems important for the implementation of #413

rupanshusoi:
The same error occurs with 
```
import "regent"
task foo(r : region(ispace(int2d), int))
  where reads writes(r) do
end
task main()
  var r = region(ispace(int2d, {10, 10}), int)
  fill(r, 1)
  var is = ispace(int2d, {5, 5})
  var p = partition(equal, r, is)
  __demand(__index_launch)
  for i in is do
    foo(p[i % is.bounds])
  end
end
regentlib.start(main)
```
even after implementing the modulus operator for optimising index launches.

elliottslaughter:
Both of these no longer reproduce as of a23407b06f4d58da0d5f98626cf15c2acc2ae948.

PR #827, opened 2020-04-27
Regent: fix for ispace.bounds in loop expression

rupanshusoi:
Fix for #817


language/src/regent/codegen.t
```diff
@@ -2956,11 +2972,12 @@ local function make_partition_projection_functor(cx, expr, loop_index, color_spa
       [value.actions];
     end)
 
-    local terra partition_functor(runtime : c.legion_runtime_t,
+    local terra partition_functor([cx.runtime],
                                   mappable : c.legion_mappable_t,
                                   idx : uint,
                                   parent : c.legion_logical_partition_t,
                                   [point])
+      var runtime = [cx.runtime]
```
elliottslaughter, line 2980: I know this shortens the diff, but it would probably make the final code cleaner if we use `[cx.runtime]` below.

language/src/regent/codegen.t
```diff
@@ -2948,6 +2948,22 @@ local function make_partition_projection_functor(cx, expr, loop_index, color_spa
...
   end
 
+  -- Hack: Rooting any ispaces present manually
+  if free_vars then
+    for _, symbol in ipairs(free_vars) do
+      local symbol_type = symbol:gettype()
+      if cx:has_ispace(symbol_type) then
+        local ispace = cx:ispace(symbol_type)
+        local bounds_actions
+        bounds_actions, ispace.domain, ispace.bounds = index_space_bounds(cx, ispace.index_space, symbol_type)
```
elliottslaughter, line 2958: I think this could potentially mess up the parent's context, since we're setting `.domain` and `.bounds` to new symbols. In fact it should crash on any test that does two launches like this in a row.

Instead of doing this, I think we can pass `domain` and `bounds` as optional arguments, and create new symbols inside of `index_space_bounds` only in the case where they come in `nil`.

rupanshusoi: Right, this fails if I have two copies of the same loop. However, `domain` and `bounds` seem to be `nil` both times.

rupanshusoi: Right, this fails if we have two copies of the same loop. However `domain` and `bounds` appear to be `nil` even for the second loop (to be precise, `ispace.bounds.lo == nil` etc). I think this is because we are only emitting code to define them and not actually defining them inside the compiler. So right now I see no direct way to tell if the domain and bounds of an ispace have been already defined.

rupanshusoi: > Instead of doing this, I think we can pass `domain` and `bounds` as optional arguments, and create new symbols inside of `index_space_bounds` only in the case where they come in `nil`.

I think what I'm unsure of is how to decide whether to create new symbols for domain and bounds or not. What property/field do I check? All `index_space_bounds` does is assign new symbols and emit code to define them. I don't see a way to tell if we already called `index_space_bounds` for a particular ispace earlier during compilation.

> I wasn't completely sure I understood the point about `domain` and `bounds` being `nil`, if I'm not mistaken they should be already set to be symbols (set at the time initialize the index space). Is that not true?

Yes, they are already set to symbols of the correct types. I mean that if I check, say `ispace.bounds.lo`, then it is `nil` inside the compiler. (We have just emitted code to define it in `free_vars_setup`.)

elliottslaughter: > I don't see a way to tell if we already called index_space_bounds for a particular ispace earlier during compilation.

The compiler should always call `index_space_bounds` at the point where an index space is created. This means at the point where we're looking at it, I don't think there should ever be a case where it hasn't been called.

I think it's helpful to separate two distinct issues:

1. creating the symbols (i.e. `terralib.newsymbol(...)`)
2. defining the variables (i.e. `quote var [ispace.bounds] = ... end`)

We don't need (and don't want) to do (1) over again. What we need is only (2).

Currently `index_space_bounds` has both jobs, but I'm suggesting we add some additional arguments so that we make it do only (2) on the 2nd time it's called.

Make sense?

rupanshusoi: Understood. Please review, it should be fixed now. 

The confusion was that I had originally thought that we wouldn't need to define the variables again if we have a second copy of the same loop. But I now see that since we're using a projection functor to close over an index space in the loop expression, this variable definition is necessary for every loop.

elliottslaughter:
Looks good, will merge this shortly.

elliottslaughter:
Github doesn't see it, but this has been merged as of 9624aab7f03995dfea533a58c8e72ab4690a51de. Thanks!

Issue #849, opened 2020-05-21
Add support to free a processor group


gabrieltanase42:
In our usage of realm we need to to often create and destroy processor groups. Think a database where for a query request I create a processor group, execute query in that processor group and then destroy the processor group. The processor group sizes are arbitrarily and depend on the query needs and the load on other cores.

gabrieltanase42:
I already have the fix for this feature but I don;t know how to contribute it back. could you please help me with some quick instructions on how I can push my branch with changes to your repository ?

I patched on top of master branch;

gabrieltanase42:
[free_group_diff.txt](https://github.com/StanfordLegion/legion/files/4663342/free_group_diff.txt)

gabrieltanase42:
I was able to fork , clone and push my diff. I hope I did the pull request the proper way :)
Pls let me know if there are any issues cherry picking my commit

streichler:
I have created a `free_group` fork in our gitlab repo here: https://gitlab.com/StanfordLegion/legion/-/tree/free_group

My first tweak to the changes was to make the API a little more consistent:
1. Create a `ProcessorGroup` handle that's a subclass of `Processor` (just like `UserEvent` is a subclass of `Event`)
2. Changed the creation method from `Processor::create_group` to `ProcessorGroup::create_group`, leaving the old one as a deprecated wrapper.
3. Changed the proposed `Processor::free_group` method to `ProcessorGroup::destroy`.

Some more changes are needed behind the scenes to deal with deferred destruction and the like, but this should be it for externally-visible changes.  @gabrieltanase42 can you let me know if these API tweaks are going to cause major pain?

@lightsighter I tweaked the Legion code to use the non-deprecated creation method, but did not try to preserve the `ProcessorGroup` type, which will be needed if you want to destroy the groups later.  Would you like me to take a shot at that, or is it easier for you to make those changes?

lightsighter:
> Would you like me to take a shot at that, or is it easier for you to make those changes?

@streichler How are you proposing to change the mapper interface to destroy processor groups?

streichler:
>  How are you proposing to change the mapper interface to destroy processor groups?

I wasn't really.  :)  If I poked at it, it'd just be to remember the groups as `ProcessorGroup` instead of `Processor` so that they can all be cleaned up at shutdown time.

lightsighter:
[...]

PR #850, opened 2020-05-21
added support to free resources associated with a processor group; th…

gabrieltanase42:
…is is usefull in applications where we create/destroy lots of processors groups

### DIFF
--- a/runtime/realm/proc_impl.cc
+++ b/runtime/realm/proc_impl.cc
@@ -99,6 +99,16 @@ namespace Realm {
       return Processor::NO_PROC;
     }
 
+    /*static*/ void Processor::free_group(const Processor& pg)
+    {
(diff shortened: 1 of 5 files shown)


gabrieltanase42:
This is the fix to address:
https://github.com/StanfordLegion/legion/issues/849

lightsighter:
One thing that might be good would be to move this over to a merge request on gitlab since our github repo is just a clone, so all code merges are performed over there.

streichler:
> One thing that might be good would be to move this over to a merge request on gitlab since our github repo is just a clone, so all code merges are performed over there.

No need.  I'm going to create a "real" branch for this because it needs a few tweaks before it can go in.

streichler:
Closing this as it has been incorporated as part of the changes discussed in #849.

Issue #867, opened 2020-06-05
Regent: Support for terra bit operations


LonelyCat124:
Legion version: 467ddee
llvm 3.8

Regent doesn't currently appear to support some terra bitwise operations inside `task` functions:
```
import "regent"

local c = regentlib.c

local format = require("std/format")


task main()

  var one : int64 = 1LL
  var two : int64 = 2LL

  var one_xor_two : int64 = one ^ two
  var one_or_two : int64 = one or two
  var one_and_two : int64 = one and two
  var one_lshift_one : int64 = one << 1
  var two_rshift_one : int64 = two >> 1

end

regentlib.start(main)
```

I commented out the bitwise statements one by one for the following outputs:
All:
```
bash-4.2$ legion/language/regent.py test_bitwise.rg
test_bitwise.rg:13: unexpected token in expression near '^'
```

Commented out XOR:
```
bash-4.2$ legion/language/regent.py test_bitwise.rg
test_bitwise.rg:16: unexpected token in expression near '<<'
```

Commented out xor, and lshift
```
bash-4.2$ legion/language/regent.py test_bitwise.rg
test_bitwise.rg:17: unexpected token in expression near '>>'
```

Commented out xor, lshift and rshift
Code compiles - adding print statements for remaining variables:
```
bash-4.2$ legion/language/regent.py test_bitwise.rg
one or two = 3
one and two = 0
```

It looks like bitwise and and bitwise or are working correctly, but I don't think regent is parsing xor or shift operations in task code (wheras in terra blocks they work fine for me). I'm not sure if they collide with symbols used for other things in Regent - (I didn't find anything in the docs).

elliottslaughter:
Yes, we should add these.

rupanshusoi:
@elliottslaughter I've added support for these bit operations in a PR, please check.

elliottslaughter:
This has been merged with #870. Thanks @rupanshusoi !

PR #870, opened 2020-06-11
Regent: support for terra bit operations

rupanshusoi:
Fix for #867. I've tried to replicate the order of operations from C (which is what the intent already seems to be) but it does not always match with Terra. For example, `5 ^ 3 << 2` evaluates to `9` in Regent (with this PR) and C, but `24` in Terra.


elliottslaughter:
It's unfortunate that Terra and C don't agree, I didn't realize that. In the past we've chosen to match Terra rather than C. But in this case it may be better to just throw an error when there's an inconsistency (i.e. the user has a `^` and a `<<` without parens to disambiguate). At least that way you won't get caught by surprise.

@streichler and I have been talking and we think it's possible to do this by marking all AST nodes surrounded by `()` with a special attribute (e.g. `has_parens`) and then we can check later on whether a `^` has as a child a `<<` with `has_parens == false` or vice versa.

Does that make sense?

rupanshusoi:
Right, although I think I'll need some help with the implementation.

elliottslaughter:
Step 1 would be to add the `has_parens` field to the unspecialized expr AST: https://github.com/StanfordLegion/legion/blob/master/language/src/regent/ast.t#L168

After that you have to modify the parser so that it sets `has_parens` to `false` for all AST nodes by default. (This is probably the most tedious step.)

At that point, you can modify the parser for `()` expressions. Instead of just returning `expr` you can return `expr` with `has_parens` set to true: https://github.com/StanfordLegion/legion/blob/master/language/src/regent/parser.t#L492

Then, finally you can write the check. This could go in a couple of places, but probably most straightforward would be to put it the parser itself in `expr_binary_left`: https://github.com/StanfordLegion/legion/blob/master/language/src/regent/parser.t#L1590

Let me know if you have any questions.

rupanshusoi:
@elliottslaughter Please review.

elliottslaughter:
This has been merged with aded2409d389ad0f72201c1da85193d64ac3d014. Thanks!

PR #887, opened 2020-07-07
RFC: accelerator module

powderluv:
Posting this PR as an RFC for an Accelerator Module for Realm. The Accelerator could be a local FPGA or a remote node accessible via RPC (think TPU etc).  

Please let us know your thoughts and we can revise with your feedback.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,6 +234,15 @@ if(Legion_USE_CUDA)
 endif()
 
 #------------------------------------------------------------------------------#
+# Accelerator configuration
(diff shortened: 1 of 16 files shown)


lightsighter:
We'll probably move this over to gitlab for the actual merge, but you can leave it here for now. Assigning @streichler as the primary reviewer since it's mostly Realm related. I looked at the Legion changes and I have no problems with them.

powderluv:
Thanks @lightsighter 

@streichler pls let us know your thoughts. We can recreate on gitlab once we address your comments. Thanks

PR #889, opened 2020-07-08
Regent: dynamic check for index launches

rupanshusoi:
This PR adds support for emitting a dynamic O(n) check to determine the safety of an index launch when the optimiser is unable to guarantee safety statically.

Edit: Due to a merge conflict, this PR has been moved to #1054

### DIFF
--- a/language/src/regent/optimize_index_launches.t
+++ b/language/src/regent/optimize_index_launches.t
@@ -1138,10 +1138,9 @@ local function optimize_loop_body(cx, node, report_pass, report_fail)
           local passed = analyze_noninterference_self(
             loop_cx, task, arg, partition_type, mapping, loop_vars)
           if not passed then
-            -- Revert later!
-            ignore(call, "loop optimization failed: argument " .. tostring(i) ..
+            ignore(call, "static loop optimization failed: argument " .. tostring(i) ..
                 " interferes with itself")
-            --return
+            return
           end
         end
       end
@@ -1168,12 +1167,12 @@ local function optimize_loop_body(cx, node, report_pass, report_fail)
   }
 end
 
--- Initalize bitmask to false
+-- Initialize bitmask to false
 function init_bitmask(bitmask)
   local stats = terralib.newlist()
 
   local idx = std.newsymbol(int32, "i")
-  local bitmask_i = util.mk_expr_index_access(util.mk_expr_id(bitmask, std.rawref(&bitmask:gettype())), util.mk_expr_id(idx), std.rawref(&bool))
+  local bitmask_i = util.mk_expr_index_access(util.mk_expr_id_rawref(bitmask), util.mk_expr_id(idx), std.rawref(&bool))
   local assign = util.mk_stat_assignment(bitmask_i, util.mk_expr_constant(false, bool))
   stats:insert(assign)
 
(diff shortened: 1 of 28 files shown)


language/src/regent/optimize_index_launches.t
```diff
@@ -472,6 +474,9 @@ local function analyze_noninterference_self(
     if analyze_index_noninterference_self(index, cx, loop_vars)
     then
       return true
+    else
+      local needs_dynamic_check = true
```
elliottslaughter, line 478: What's the reason for the local variable here?

language/src/regent/optimize_index_launches.t
```diff
@@ -472,6 +474,9 @@ local function analyze_noninterference_self(
     if analyze_index_noninterference_self(index, cx, loop_vars)
     then
       return true
```
elliottslaughter, line 476: If we're going to make this function return two values, should we return a second one here? (Certain things can go rather strangely when unpacking `nil` values, so `false` would be a bit safer.)

language/src/regent/optimize_index_launches.t
```diff
@@ -1154,18 +1163,127 @@ local function optimize_loop_body(cx, node, report_pass, report_fail)
...
     reduce_op = reduce_op,
     args_provably = args_provably,
     free_variables = free_vars,
-    loop_variables = loop_vars
+    loop_variables = loop_vars,
+    needs_dynamic_check = needs_dynamic_check
   }
 end
 
+function init_bitmask_false(bitmask, volume)
```
elliottslaughter, line 1184: These should all be `local function`s.

language/src/regent/optimize_index_launches.t
```diff
@@ -1154,18 +1163,127 @@ local function optimize_loop_body(cx, node, report_pass, report_fail)
...
+  local stats = terralib.newlist()
+
+  -- Compute value = index_expr(i)
+  preamble:map(function(stat) stats:insert(stat) end)
+  stats:insert(util.mk_stat_assignment(util.mk_expr_id_rawref(value), index_expr))
+
+  local then_block = terralib.newlist()
+
+  local bitmask_value = util.mk_expr_index_access(util.mk_expr_id(bitmask), util.mk_expr_id(value), std.rawref(&bool))
+  local conflict_assign = util.mk_stat_assignment(util.mk_expr_id_rawref(conflict), bitmask_value)
```
elliottslaughter, line 1209: This is a minor preference, but since the variable is used exactly once on the following line, I'd be inclined not to create a variable and to instead just pass it straight to the `:insert()` call.

language/src/regent/optimize_index_launches.t
```diff
@@ -1154,18 +1163,127 @@ local function optimize_loop_body(cx, node, report_pass, report_fail)
...
+  stats:insert(bounds_check)
+
+  return stats
+end
+
+function insert_dynamic_check(index_launch_ast, unoptimized_loop_ast)
+  local stats = terralib.newlist()
+
+  -- Optimize only if we are indexing into a partition
+  local proceed, p_symbol = pcall(function() return index_launch_ast.call.args[1].value.value end)
```
elliottslaughter, line 1230: This looks like a hard-coded check for a specific argument?

Also, if this is truly a requirement on the dynamic optimization, it should probably be lifted into the checking phase of the optimization so that we can issue an error message if the loop cannot be dynamically optimized. (If it's just an expedience to make the initial code faster to write, let's generalize it instead.)

Also since we're not doing things with "opaque" objects (like Terra ASTs) here, I'd be inclined to write out the check rather than use `pcall`. I generally reserve `pcall` for cases where such a check *cannot* reasonably be written out (e.g., like I said, when checking whether a Terra AST type checks, which we can't really do by any means other than invoking Terra to compile it).

language/tests/regent/compile_fail/optimize_index_launch_list6.rg
```diff
@@ -12,6 +12,9 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 
+-- runs-with:
+-- [["-findex-launch-dynamic", "0"]]
```
elliottslaughter, line 16: Is there a reason for these tests to have the dynamic check disabled?

elliottslaughter (commented):
Overall looks like a good first step. Looking forward to seeing the rest of it come together.

language/src/regent/optimize_index_launches.t
rupanshusoi: Fixed.

language/tests/regent/compile_fail/optimize_index_launch_list6.rg
rupanshusoi: Yes, if we remove the flag then these tests will no longer fail (note that they are in `compile_fail`).

language/src/regent/optimize_index_launches.t
rupanshusoi: I had hoped this would improve readability when we return `false, true` on the next line, but I'm not confident that this is the case.

language/src/regent/optimize_index_launches.t
rupanshusoi: Fixed.

language/src/regent/optimize_index_launches.t
rupanshusoi: Fixed.

language/src/regent/optimize_index_launches.t
rupanshusoi: Ok, I've replaced this with a somewhat long check that uses the compiler's `std`. I don't think we can put this anywhere else because we only generate the check for partitions and the task might have other, non-partition arguments which we need to skip.

PR #908, opened 2020-08-04
Fix a void* cast in cudart with CUDA11

powderluv:
You run into:legion/runtime/realm/cuda/cudart_hijack.cc:436:22: error:
invalid conversion from ‘const void*’ to ‘void*’ [-fpermissive]
  attr->hostPointer = (const void*) ptr;
                      ^~~~~~~~~~~~~~~~~

TEST: Verify legion builds
Change-Id: Ia15247b728d85860154446b6cffcad1bc868827b

### DIFF
--- a/runtime/realm/cuda/cudart_hijack.cc
+++ b/runtime/realm/cuda/cudart_hijack.cc
@@ -433,7 +433,7 @@ extern "C" {
 	attr->type = cudaMemoryTypeUnregistered;
 	attr->device = 0;
 	attr->devicePointer = 0;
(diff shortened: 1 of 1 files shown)


streichler:
Yep, this looks like the right way to fix this.  I'll double-check on my system and then merge.  Thanks!

streichler:
I ended up spelling it slightly differently, but this should be fixed in the master branch with commit 1ad3e22.  Thanks again for catching this!

powderluv:
Looks great. Thanks.

PR #919, opened 2020-08-26
update README to include variables for MPI

rzambre:
Setting USE_MPI=1 alone does not configure Legion to use the MPI
backend. It must be coupled with REALM_NETWORKS=mpi.


rzambre:
The README currently does not mention the existence of the REALM_NETWORKS variable. That documentation is hidden in the CHANGES.txt file.

streichler:
I agree that the README should probably talk about MPI network layer, but that's not what `USE_MPI=1` does.  `USE_MPI=1` enables the MPI compatibility logic in Legion and Realm.  Is that what you're trying to get, or did you actually want the MPI
network layer in Realm?

rzambre:
I see. I want to actually use the MPI backend in Realm.
Is it possible to use the MPI network layer with just `REALM_NETWORKS=mpi` and without `USE_MPI=1` ?

streichler:
> Is it possible to use the MPI network layer with just `REALM_NETWORKS=mpi` and without `USE_MPI=1` ?

Yes, `REALM_NETWORKS=mpi` should be sufficient on its own.

rzambre:
Got it, I have updated the PR to reflect the availability of the `REALM_NETWORKS` variable

streichler:
Thanks.  Rather than taking this little change as is, I'm inclined to use this PR as a forcing function to do a larger update of those docs to include all the other build switches we've been bad at documenting over the years.  :)

rzambre:
Sounds great :)

Issue #943, opened 2020-10-05
Regent - Assertion failure when partitioning 1D region into 3D ispace using equal partition.


LonelyCat124:
I'm not sure if this is a Legion bug or Regent allowing me to do something thats not allowed. The following code fails an assertion:
```
import "regent"
fspace part{
  thing : int,
  thing2 : int
}
task main()
  var i = 2
  var space = ispace(int1d, {1000})
  var thing = region(space, part)
  var part = partition(equal, thing, ispace(int3d,{5,5,5}))
end

regentlib.start(main)
```

```
terra: /home/aidan/Legion/legion/runtime/legion/legion_domain.inl:1061: Legion::Domain::operator Legion::Rect<DIM, T>() const [with int DIM = 1; T = long long int; Legion::Rect<DIM, T> = Realm::Rect<1, long long int>]: Assertion `DIM == dim' failed.
```

 I'm not sure if I'm just not allowed to partition a 1D region using a 3D index space, but I know if I have an `int3d` inside the field space, I can partition with:
```
  fill(thing.int3d_parameter, int3d({0,0,0}))
  var part = partition(thing.int3d_parameter, ispace(int3d, {5,5,5}))
```
with no issues.

I tried to grab a backtrace but even building regent with `--debug` this was the best I could get, not quite clear why
```
  [0] = /lib/x86_64-linux-gnu/libpthread.so.0(+0x12890) [0x7f314a192890]
  [1] = /lib/x86_64-linux-gnu/libc.so.6(gsignal+0xc7) [0x7f314901ee97]
  [2] = /lib/x86_64-linux-gnu/libc.so.6(abort+0x141) [0x7f3149020801]
  [3] = /lib/x86_64-linux-gnu/libc.so.6(+0x3039a) [0x7f314901039a]
  [4] = /lib/x86_64-linux-gnu/libc.so.6(+0x30412) [0x7f3149010412]
  [5] = /home/aidan/Legion/legion/bindings/regent/libregent.so(_ZNK6Legion6DomaincvN5Realm4RectIXT_ET0_EEILi1ExEEv+0x47) [0x7f3146e9c90d]
  [6] = /home/aidan/Legion/legion/bindings/regent/libregent.so(legion_domain_get_rect_1d+0x4b) [0x7f3146ef78a0]
  [7] = [0x7f314aa35454]
  [8] = [0x7f314aa34cdb]
[...]

PR #945, opened 2020-10-07
Regent: fix bug in equal partition codegen

rupanshusoi:
Fix for #943

### DIFF
--- a/language/src/regent/codegen.t
+++ b/language/src/regent/codegen.t
@@ -4587,7 +4587,6 @@ function codegen.expr_partition_equal(cx, node)
     end
   else
     local dim = region_type:ispace().dim
-    local colors_dim = colors_type.dim
     local transform = terralib.newsymbol(
       c["legion_transform_" .. tostring(dim) .. "x" .. tostring(dim) .. "_t"], "transform")
     local extent = terralib.newsymbol(
@@ -4595,7 +4594,7 @@ function codegen.expr_partition_equal(cx, node)
     local domain_get_rect =
       c["legion_domain_get_rect_" .. tostring(dim) .. "d"]
     local colors_domain_get_rect =
-      c["legion_domain_get_rect_" .. tostring(colors_dim) .. "d"]
+      c["legion_domain_get_rect_" .. tostring(colors_type.dim) .. "d"]
     local create_domain_transform =
       c["legion_domain_transform_from_" .. tostring(dim) .. "x" .. tostring(dim)]
     local create_domain =
--- a/language/src/regent/codegen.t
+++ b/language/src/regent/codegen.t
@@ -4587,12 +4587,15 @@ function codegen.expr_partition_equal(cx, node)
     end
   else
     local dim = region_type:ispace().dim
+    local colors_dim = colors_type.dim
     local transform = terralib.newsymbol(
       c["legion_transform_" .. tostring(dim) .. "x" .. tostring(dim) .. "_t"], "transform")
     local extent = terralib.newsymbol(
       c["legion_rect_" .. tostring(dim) .. "d_t"], "extent")
     local domain_get_rect =
       c["legion_domain_get_rect_" .. tostring(dim) .. "d"]
+    local colors_domain_get_rect =
+      c["legion_domain_get_rect_" .. tostring(colors_dim) .. "d"]
     local create_domain_transform =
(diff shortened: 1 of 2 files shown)


elliottslaughter:
Would it be possible to get a test case that verifies the fix (i.e. fails without it, and passes with it)?

Otherwise everything looks good.

Thanks.

rupanshusoi:
@elliottslaughter Done.

elliottslaughter:
Actually this test fails for me. After a bit of digging, it looks like Regent is actually using two different code paths in the compiler. You can see this with a slightly modified test:

```
import "regent"

task main()
  -- Case 1. unstructured partitioned by 3D
  var r = region(ispace(ptr, 10), int)
  var p = partition(equal, r, ispace(int3d, {5, 5, 5}))

  -- Case 2. 1D partitioned by 3D
  var s = region(ispace(int1d, 10), int)
  var q = partition(equal, s, ispace(int3d, {5, 5, 5}))
end
regentlib.start(main)
```

Case 1 passes (with the patch) but case 2 fails.

It seems that case 2 is failing because we tried to set up a custom code path to make the equal partition output more sane in the case of multi-dimensional regions. But I'm not honestly sure if that makes sense when the region and color space dimensionality don't match. So maybe we should just fall back to the runtime equal partition in both cases.

rupanshusoi:
Hmm, this test does not fail for me. I'm using an older commit, but for Regent it should be equivalent to the latest master.

rupanshusoi:
@elliottslaughter I tested with the latest master, and your test did **not** fail for me. I do see CI failing for the new test I added (again, which seems to works fine for me), so not completely sure what's going on.

elliottslaughter:
I went back to your original commit (on the theory that the merge might have gone bad), but it's still failing for me.

Do you have Sapling access? That's where I'm able to reproduce this.

rupanshusoi:
@elliottslaughter Something weird is happening. I made a fresh clone on Sapling, built the latest master and applied this PR with `git pull origin pull/945/head`. The snippet you posted above did **not** fail for me.

> Case 1 passes (with the patch) but case 2 fails.

I think the changes in this PR aren't being applied for you (and for CI), because case 1 should work without this PR as well.

Can you please apply the patch I've attached and check if your code snippet still fails? I'm confident that it will work.
[patch.txt](https://github.com/StanfordLegion/legion/files/5496674/patch.txt)

elliottslaughter:
I have rebased and pushed this to branch `regent-pr-945` at Gitlab. If it passes CI there then I'm inclined to merge it.

elliottslaughter:
You can watch the tests run at https://gitlab.com/StanfordLegion/legion/-/pipelines/245762333/builds

elliottslaughter:
I'm still getting a failure in CI, see e.g.: https://gitlab.com/StanfordLegion/legion/-/jobs/981724141

lightsighter:
This is almost certainly a compiler/application bug. It's saying the type of the index space data passed in does not match the type of the index space handle.

rupanshusoi:
I'm not sure if this is worth pointing out, but I could only get the error to reproduce on Sapling on installing with `--spy`. I can see similar behaviour with CI, where it passed with Spy disabled: https://gitlab.com/StanfordLegion/legion/-/jobs/981724142

Does this tell us anything useful?

lightsighter:
I can reproduce the same failure after building Regent without Legion Spy support. Did you make sure you built Legion in debug mode with `--debug`? You won't hit this assertion in a release mode because Legion removes all assertions in release mode. Looking at the assertion below though, this is definitely the compiler's fault.

The part I don't understand is why Regent is invoking 'legion_create_partition_by_restriction' at all since this is a request to create an equal partition. If you're going to call `legion_create_partition_by_restriction` at all, then the second dimension of the transform must have the same dimensionality as the color space (in this case 3). However, the transform being passed into 'legion_create_partition_by_restriction` here has dimensionality 1x1. It should really be 1x3.

elliottslaughter:
I pushed a rebase and final fix to https://gitlab.com/StanfordLegion/legion/-/tree/regent-pr-945. This will hopefully take care of it. While reading the code, I realized that what we had really only would work if `dim == color_dim`, because it was assuming we could block out each dimension one at a time. So now we only follow that code path if that's going to work, and fall back to the runtime otherwise.

elliottslaughter:
This has been merged. Thanks!

PR #970, opened 2020-11-23
Pygion projection functor

ritabt:


magnatelee:
@ritabt could you make this PR based on either master or control_replication only? This PR seems to be mixing changes from two different runtime branches, which isn't really a great idea. Also, it makes it really hard for us to review what's really changed here.

elliottslaughter:
I thought I changed the base, but maybe Github is failing to reanalyze it. Stay tune....

PR #1097, opened 2021-06-08
Regent: dynamic projection functor cross-checks

rupanshusoi:
This PR extends the work in #1054 by adding support for projection functor cross-checks, i.e., checking multiple arguments of the same partition but using different projection functors against each other. For `N` arguments of the same partition, the check takes `O(N)` time.

For example, the safety of the following index launch can now be dynamically checked:

```
__demand(__index_launch)
for i in ... do
  foo(p[i % a], p[i % b], p[i % c])
end
```
Additionally, this PR also adds support for analyzing projection functors of dimension up to 9.

### DIFF
--- a/language/src/regent/optimize_index_launches.t
+++ b/language/src/regent/optimize_index_launches.t
@@ -43,9 +43,9 @@ end
 
 local function get_partition_dim(expr)
   if expr:is(ast.typed.expr.Projection) then
-    return std.as_read(expr.region.value.expr_type).colors_symbol.symbol_type.dim
+    return std.as_read(expr.region.value.expr_type):colors().dim
   else
-    return std.as_read(expr.value.expr_type).colors_symbol.symbol_type.dim
+    return std.as_read(expr.value.expr_type):colors().dim
   end
 end
 
@@ -267,11 +267,10 @@ local function analyze_noninterference_previous(
     end
   end
 
-  if #args_interfering == 0 then
+  if next(args_interfering) == nil then
     return true
-  else
-    return false, args_interfering
   end
+  return false, args_interfering
 end
 
 local function add_exprs(lhs, rhs, sign)
@@ -1332,24 +1331,19 @@ local function collect_and_sort_args(args, call_args, task, param_region_types)
   return array
 end
 
-local c = terralib.includecstring([[
-#include <stdio.h>
-#include <stdlib.h>
-]])
-
 terra throw_dynamic_check_error(loc : rawstring, arg1 : uint8, arg2 : uint8)
-  var stderr = c.fdopen(2, "w")
+  var stderr = std.c.fdopen(2, "w")
   if arg1 == arg2 then
-    c.fprintf(stderr, [[Errors reported during runtime.
+    std.c.fprintf(stderr, [[Errors reported during runtime.
 %s: loop optimization failed: argument %d interferes with itself
 ]], loc, arg1)
   else
-    c.fprintf(stderr, [[Errors reported during runtime.
+    std.c.fprintf(stderr, [[Errors reported during runtime.
 %s: loop optimization failed: argument %d interferes with argument %d
 ]], loc, arg1, arg2)
   end
-  c.fflush(stderr)
-  c.abort()
+  std.c.fflush(stderr)
+  std.c.abort()
 end
 
 local function insert_dynamic_check(is_demand, args_need_dynamic_check, index_launch_ast, unopt_loop_ast)
--- a/language/src/regent/optimize_index_launches.t
+++ b/language/src/regent/optimize_index_launches.t
@@ -1310,6 +1310,26 @@ local function collect_and_sort_args(args, call_args, task, param_region_types)
   return array
 end
 
+local c = terralib.includecstring([[
+#include <stdio.h>
+#include <stdlib.h>
+]])
+
+terra assert_dyn_check_error(x : bool, message : rawstring, arg1 : uint8, arg2: uint8)
+  if not x then
+    var stderr = c.fdopen(2, "w")
+    if arg1 == arg2 then
+      c.fprintf(stderr, "Errors reported during runtime.\n%s%d%s\n",
+        message, arg1, " interferes with itself")
+    else
+      c.fprintf(stderr, "Errors reported during runtime.\n%s%d%s%d\n",
+        message, arg1, " interferes with argument ", arg2)
+    end
+    c.fflush(stderr)
+    c.abort()
+  end
+end
+
 local function insert_dynamic_check(is_demand, args_need_dynamic_check, index_launch_ast, unopt_loop_ast)
   local check = terralib.newlist()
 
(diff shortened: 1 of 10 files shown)


elliottslaughter:
Just FYI, we had CI fail on this one: https://gitlab.com/StanfordLegion/legion/-/pipelines/317130695/builds

language/src/regent/optimize_index_launches.t
```diff
@@ -1157,13 +1164,13 @@ local function optimize_loop_body(cx, node, report_pass, report_fail)
...
               for _, failure in pairs(failures_i) do
-                if arg.value.value == args[failure].value.value then
+                if get_partition_symbol(arg) == get_partition_symbol(args[failure]) then
                   table.insert(args_need_dynamic_check, { i, failure })
                 end
               end
             else
               report_fail(call, "loop optimization failed: argument " .. tostring(i) ..
-                  " interferes with argument " .. tostring(failure_i))
+                   " interferes with argument " .. tostring(failures_i[1]))
```
elliottslaughter, line 1181: Probably unintended to add a space here.

language/src/regent/optimize_index_launches.t
```diff
@@ -1349,11 +1356,11 @@ local function insert_dynamic_check(is_demand, args_need_dynamic_check, index_la
   for _, args in pairs(collected_args) do
     local stats = terralib.newlist()
 
-    local dim = call_args[args[1]].expr_type:ispace().dim
+    local dim = std.as_read(call_args[args[1]].value.expr_type).colors_symbol.symbol_type.dim
```
elliottslaughter: I think you can replace `.colors_symbol.symbol_type` with `:colors()`.

language/src/regent/optimize_index_launches.t
```diff
@@ -41,6 +41,14 @@ local function get_partition_symbol(expr)
   return expr.value.value
 end
 
+local function get_partition_dim(expr)
+  if expr:is(ast.typed.expr.Projection) then
+    return std.as_read(expr.region.value.expr_type).colors_symbol.symbol_type.dim
```
elliottslaughter, line 46: Same here.

language/src/regent/optimize_index_launches.t
```diff
@@ -41,6 +41,14 @@ local function get_partition_symbol(expr)
...
+  else
+    return std.as_read(expr.value.expr_type).colors_symbol.symbol_type.dim
```
elliottslaughter, line 48: Again.

language/src/regent/optimize_index_launches.t
elliottslaughter: This is outdated since I was looking at an old commit, but the same idea applies to the new patch in a different location.

language/src/regent/optimize_index_launches.t
```diff
@@ -247,10 +263,15 @@ local function analyze_noninterference_previous(
         -- Index non-interference is handled at the type checker level
         -- and is captured in the constraints.
     then
-      return false, i
+      table.insert(args_interfering, i)
     end
   end
-  return true
+
+  if #args_interfering == 0 then
```
elliottslaughter, line 270: I'm not sure `#` can be trusted if you're going to potentially have `nil` entries in the list. See: https://www.lua.org/manual/5.1/manual.html#2.5.5

I think you actually need to iterate this with `pairs` and break out of the loop if you find something.

language/src/regent/optimize_index_launches.t
```diff
@@ -1146,12 +1167,20 @@ local function optimize_loop_body(cx, node, report_pass, report_fail)
...
-            return
+            if emit_dynamic_check then
+              for _, failure in pairs(failures_i) do
+                if get_partition_symbol(arg) == get_partition_symbol(args[failure]) then
+                  table.insert(args_need_dynamic_check, { i, failure })
+                end
+              end
+            else
+              report_fail(call, "loop optimization failed: argument " .. tostring(i) ..
+                   " interferes with argument " .. tostring(failures_i[1]))
```
elliottslaughter, line 1181: I made a previous comment on this line, please ignore it.

language/src/regent/optimize_index_launches.t
```diff
@@ -1219,140 +1248,242 @@ local function init_bitmask(bitmask, volume)
...
-       std.is_region(arg.expr_type) and
-       std.is_partition(std.as_read(arg.value.expr_type)) then
+  return array
+end
 
-      local dim = arg.expr_type:ispace().dim
+local c = terralib.includecstring([[
+#include <stdio.h>
+#include <stdlib.h>
+]])
```
elliottslaughter, line 1334: Instead of including C headers in the middle of this file, let's use `std.c`, which I think already includes a superset of these headers.

elliottslaughter:
Looks like there are some line numbers/error messages off in the `compile_fail` tests:

https://gitlab.com/StanfordLegion/legion/-/jobs/1436334406

language/src/regent/optimize_index_launches.t
rupanshusoi: Done.

language/src/regent/optimize_index_launches.t
rupanshusoi: Fixed.

language/src/regent/optimize_index_launches.t
rupanshusoi: I replaced it with `next(args_interefering) == nil` which might be more efficient.

elliottslaughter:
Running now at https://gitlab.com/StanfordLegion/legion/-/pipelines/340170820

elliottslaughter:
Let me try a rebase, it may just be missing some commits on master.

elliottslaughter:
Running as https://gitlab.com/StanfordLegion/legion/-/pipelines/340204733 . Note I pushed the rebase to upstream.

language/src/regent/optimize_index_launches.t
```diff
@@ -247,10 +263,14 @@ local function analyze_noninterference_previous(
         -- Index non-interference is handled at the type checker level
         -- and is captured in the constraints.
     then
-      return false, i
+      table.insert(args_interfering, i)
```
elliottslaughter, line 266: If you make `args_interfering` into a `terralib.newlist()` then you can do `args_interfering:insert(i)` here instead of needing to explicitly invoke `table.insert`.

elliottslaughter (approved):
Looks good, just found one minor code cleanup to do and then we can merge it. Thanks!

language/src/regent/optimize_index_launches.t
rupanshusoi: I'm attaching the potential changes as a patch file. The two issues are that i) I still use `table.insert` in other places, and ii) I couldn't find a nicer way than using `:exists(...)` to check if a Terra list is empty.
[patch.txt](https://github.com/StanfordLegion/legion/files/6856922/patch.txt)

elliottslaughter: A Terra list is still a Lua table with keys starting at 1. So the other code here doesn't need to change, you can still use `ipairs` or `pairs` to iterate it, and so on. Of course if you prefer the functional style you're welcome to use `:map` if you want to, but it's not necessary.

rupanshusoi: Ok, thanks, that makes sense. I just pushed this. I stuck to using `pairs` for consistency.

elliottslaughter:
Merged via 9e3359851932ae24ef8e98d07c123946fd325c23. Github won't see it until the mirror sync, but it's in master on Gitlab.

Thanks again!

PR #1151, opened 2021-10-04
Regent: support cross products in index launches

rupanshusoi:
This PR adds support for analysing and generating code for cross products in an index launch. For non-interference analysis, both static and dynamic, a cross product is treated like its first partition.


language/src/regent/ast.t
```diff
@@ -666,4 +666,11 @@ function ast.mapreduce_stat_postorder(map_fn, reduce_fn, node, init)
     reduce_fn, node, init, is_stat_node)
 end
 
+function ast.get_base_indexed_node(node, previous_index)
```
elliottslaughter, line 669: This seems like the sort of thing that would be more appropriate in `ast_util.t`?

language/src/regent/codegen.t
```diff
@@ -2939,13 +2938,45 @@ local function make_partition_projection_functor(cx, expr, loop_index, color_spa
...
+  cx = cx:new_local_scope()
+
+  if expr:is(ast.typed.expr.Projection) then
+    expr = expr.region
+  end
+  assert(expr:is(ast.typed.expr.IndexAccess))
+
+  -- Never return 0 for cross products
+  if is_identity_projection_all(expr, loop_index) and
+     not std.is_cross_product(std.as_read(ast.get_base_indexed_node(expr).expr_type))
```
elliottslaughter, line 2975: Wouldn't it be more correct to return 0 iff the projection functor is exactly one level deep (and is the identity)? If we (hypothetically) supported another nested type in the future, that would avoid a need for an explosion of custom code here.

language/src/regent/codegen.t
```diff
@@ -2994,10 +3025,19 @@ local function make_partition_projection_functor(cx, expr, loop_index, color_spa
       [value.actions];
     end)
 
+    local parent = terralib.newsymbol(c.legion_logical_partition_t, "parent")
+    local index_access
+    if std.is_cross_product(std.as_read(ast.get_base_indexed_node(expr).expr_type)) then
```
elliottslaughter, line 3030: Don't we want this check to go the other way? If it's a partition then we wrap it. Cross products can fall under the "everything else" case.

language/src/regent/codegen.t
```diff
@@ -3400,7 +3426,9 @@ local function expr_call_setup_partition_arg(
 
   free_vars_setup:insertall(loop_vars_setup)
 
-  local needs_non_identity_functor = not is_identity_projection(arg_value, loop_index)
+  -- Cross products always need the full-blown partition_functor
+  local needs_non_identity_functor = (not is_identity_projection_all(arg_value, loop_index)) or
```
elliottslaughter, line 3430: See above.

language/src/regent/optimize_index_launches.t
```diff
@@ -858,6 +867,17 @@ local function analyze_is_simple_index_expression(cx, node)
     node, true)
 end
 
+-- `p` must be a partition or cross product
+local function analyze_is_loop_invariant_recursive(cx, p)
+  if not analyze_is_loop_invariant(cx, p) then
+    if std.is_cross_product(std.as_read(p.value.expr_type)) then
```
elliottslaughter, line 873: Shouldn't this be based on the AST structure not the type?

elliottslaughter (commented):
Looks good, just a few minor design issues. Thanks!

language/src/regent/ast.t
rupanshusoi: Right, thanks.

language/src/regent/optimize_index_launches.t
rupanshusoi: I don't understand what is wrong here. If we have `cp[i][i][i]` in an index launch, then we only need to ensure that the base cross product `cp` is loop invariant. Since `analyze_loop_invariant` returns `false` for intermediate cross products (`cp[i][i]`, for instance), I check whether that is the case and recursively call it on the next cross product one level up in the AST (`cp[i]`, in this case).

elliottslaughter:
Running: https://gitlab.com/StanfordLegion/legion/-/pipelines/391434588

elliottslaughter:
Also, note: if you want to reset this branch to match what I'm working on, you can pull from here: https://gitlab.com/StanfordLegion/legion/-/tree/cross-product-squash

elliottslaughter:
Update on this: HTR seems to be breaking, as best I can tell, because it overrides `list:find()` with different semantics. Following up with the HTR developers.

elliottslaughter:
I was able to confirm that HTR works when its overriding definition of `list:find()` has been removed (which HTR does not seem to use anyway).

elliottslaughter:
This has been merged in f7a0b37cfd6a5ca936bfad7417d686740c5c1473. Thanks again for all your help with this!

PR #1162, opened 2021-11-03
CMake update to expose parameters for max node and procs settings. 

pmccormick:
Added new CMake parameters to expose the build-time parameters for setting the maximum numbers of nodes and processors.  The requirement for a power-of-two of both these parameters is checked/enforced at config time. 

Apologies for the additional diff entries -- my editor was set to clean up trailing white space (a requirement for upstreaming in spack that I forgot to turn off).


streichler:
Rebased, tweaked a bit, and merged into `master` with commit 09f30796.  This'll go into `stable` before the end of the month.

PR #1174, opened 2022-01-20
nit: bump VERSION to 21.12.0

dzhang314:
Noticed that the VERSION file hadn't been updated in a while, so here's a tiny PR that updates it to the latest release, 21.12.0.


elliottslaughter:
I think this has been done on master: https://github.com/StanfordLegion/legion/blob/master/VERSION

PR #1210, opened 2022-03-08
realm: add FPGA module

dycz0fx:
Add a Realm FPGA module and two vadd examples (one Legion example in examples/fpga, one Realm example in examples/realm_fpga).


fdouglis:
Pulling for this (pun intended)

nindanaoto:
Is this PR going to merge? This seems to be what I hoped to get for a year..... (The task graph based scheduler for heterogeneous environment including FPGA)

fdouglis:
@streichler as I work on the report for our project, I note a comment from the report a year ago about how this would be merged into Legion ... it looks like it's still on the back-burner.  Any hope here?

streichler:
One of the key issues (how to test this in CI) looks like it might be more solved than I thought it was.  I see there are instructions for running the examples on an AWS node.  Are there any other cloud providers that have compatible hardware?

Another issue is the name of the module.  I think "fpga" is too generic (e.g. we didn't call the cuda module "gpu").  If this is specific to Xilinx FPGAs (or maybe even a subset of the Xilinx family?), should we have a more specific name?  Or is it mostly an OpenCL interface and might work on other device that support OpenCL?

fdouglis:
We're getting ready to submit the final report to DARPA.  It'd be so nice if this actually saw the light of day ...

streichler:
It looks like the ball is in @dycz0fx 's court right now?

elliottslaughter:
@dycz0fx is not working on this anymore, but we will probably pick up the ball here at SLAC/LANL soonish. Maybe not in time for the DARPA report.

We should assign this to @alexaiken and @pmccormick.

dycz0fx:
> One of the key issues (how to test this in CI) looks like it might be more solved than I thought it was. I see there are instructions for running the examples on an AWS node. Are there any other cloud providers that have compatible hardware?
> 
> Another issue is the name of the module. I think "fpga" is too generic (e.g. we didn't call the cuda module "gpu"). If this is specific to Xilinx FPGAs (or maybe even a subset of the Xilinx family?), should we have a more specific name? Or is it mostly an OpenCL interface and might work on other device that support OpenCL?

Yeah, FPGA might be too generic. It only works on the Xilinx FPGA with XRT (Xilinx Runtime library) support.  
I used the OpenCL extension of the XRT, but I did not try this OpenCL code on other devices that support OpenCL.

PR #1248, opened 2022-05-03
Default mapper: make default_policy_select_constraints respect an existing specialized constraint

bandokihiro:
In my use case, a compact sparse instance wasn't created because the default mapper would override the associated specialized constraint. This small change prevents this when a specialized constraint already exists.


elliottslaughter:
I took the version from #1313, which is slightly more comprehensive than this. Thanks for the patch.

PR #1362, opened 2022-12-13
Add the UCX network module

SeyedMir:


language/install.py
```diff
@@ -290,6 +290,7 @@ def install_bindings(regent_dir, legion_dir, bindings_dir, python_bindings_dir,
             os.mkdir(build_dir)
         flags = (
             ['-DCMAKE_BUILD_TYPE=%s' % ('Debug' if debug else 'Release'),
+             '-DLegion_NETWORKS=%s' % os.environ.get('REALM_NETWORKS', ''),
```
elliottslaughter, line 293: Could we do this in the style of the checks down the page from here? I.e.:

```
(['-DLegion_NETWORKS=%s' % os.environ['REALM_NETWORKS']] if 'REALM_NETWORKS' in os.environ else []) +
```

Feels appropriate to put this under `-DGASNet_CONDUIT`.

CMakeLists.txt
```diff
@@ -455,6 +455,11 @@ if(Legion_USE_GASNet)
   list(APPEND Legion_NETWORKS gasnet1)
 endif()
 
+option(Legion_USE_UCX "Enable the distributed UCX backend" OFF)
+if(Legion_USE_UCX)
+  list(APPEND Legion_NETWORKS ucx)
+endif()
+
```
elliottslaughter, line 462: I thought we were moving away from this pattern. We don't even have a `gasnetex` option right now. Just configure this via `Legion_NETWORKS`.

lightsighter:
@elliottslaughter @streichler Does it make sense to get rid of our GASNet UCX tests in the CI now that we're going to be testing UCX directly? It seems like users should never need to use the GASNet UCX conduit now that we have a direct UCX implementation.

streichler:
> @elliottslaughter @streichler Does it make sense to get rid of our GASNet UCX tests in the CI now that we're going to be testing UCX directly? It seems like users should never need to use the GASNet UCX conduit now that we have a direct UCX implementation.

In a perfect world, I'd keep them and add new tests, as the two network modules will likely be using UCX in different enough ways that the right choice of network module may still be application dependent.  In the real world where we are limited to 150 CI configs, it's likely we'll reduce the gasnet-ucx coverage (and maybe some other coverage as well) to make room for the new test configs.

language/install.py
elliottslaughter: Looks fine aside from a slight typo in the new code where a `)` has to be moved to the next line.

PR #1399, opened 2023-02-10
Fix shutil typo

SeyedMir:


lightsighter:
Thanks! We usually do merges over on gitlab since github is just a mirror of gitlab. I manually moved this one over.

PR #1445, opened 2023-03-24
Add OMP_PROC and PY_PROC cases for Realm Processor get_kind_name

brombaut:
Just came across these missing cases when I was playing around with the Realm OpenMP module. I checked and didn't find any, but please advise if there are any tests that should also be changed to cover these additions.


streichler:
@brombaut thanks for catching this.  There's a preprocessor-based way to fix this so that it automatically updates for any other kinds added in the future.  I'll put that change in a branch so that you can confirm it solves your original problem.

PR #1489, opened 2023-06-14
mem_trace updates

syamajala:
This pull request fixes a bug with `-DLEGION_TRACE_ALLOCATION` and prints RSS with allocation traces.

It also updates mem_trace to dump backtrace hashes on every allocation and uses binary files to speed up writing. There is a separate mem_trace_csv utility for parsing those binary files and converting them to csv.


lightsighter:
Can you move this over to a gitlab merge request? The tools like fine. I don't like the dependence on `sys/resource.h` that's naked in `runtime.h` because that is not portable across OS-es. In general, we should try not to do anything in Legion that is OS-specific (that is Realm's purview). It's ok if you want to guard it with a compile-time macro that users can opt into at build time if they're willing to take on the responsibility for the non-portable include.

syamajala:
I guess that include of sys/resource.h should be guarded by #ifdef LEGION_TRACE_ALLOCATION. 

Elliott also suggested including a makefile for mem_trace so I added one as well. 

I moved the pull request to gitlab.

